Issue #39, opened 2021-02-09
Create user state and change view according to user state


dmitryy90:
### Task description
Create user state and change view according to user state

PR #52, opened 2021-02-12
#39 feature/ user token and registration

litvinets:
1. Added registration and login
2. Added authorization, logout, and call API
3. Added user state
4. Added changing of header view depending on the user state
5. HeaderComponent, AppComponent have fixed tests
6. FirterCityComponent has commented tests because of the loop (seems like)


src/app/header/header.component.html
```diff
@@ -1,12 +1,21 @@
 <div class="header-cover" fxLayout="row" fxLayoutAlign="space-between">
   <div>Позашкілля</div>
-  <div *ngIf="isAuthorized; else isUnauthorized">
-    Увійти
-  </div>
+  <ng-container
+    *ngIf="(isAuthorized$ | async) then YesAuthorized; else NotAuthorized">
```
dmitryy90, line 4: is it more meaningful to use authorized and unauthorized instead of yes no prefixes?

dmitryy90 (approved):

Issue #54, opened 2021-02-12
Add error handler for isAuthorized method


dmitryy90:
### Task description
Add error handler for isAuthorized method

### To do 
1. - [ ] Make app working even there is error in open id service and show notification

PR #56, opened 2021-02-13
#54 error handler for is authorized

litvinets:
1. In the commit "error handler" you can find only this solution 
 .pipe(
        tap(response => {
          return response;
        }),
        catchError(err => {
          console.log("Error is"+ err);
          return throwError(err);
        })
      )
2. But then I've decided to add http-error.interceptor.ts and it reacts to all HTTP errors


dmitryy90 (changes requested):

src/app/shared/modals/registration/registration.module.ts
```diff
@@ -6,19 +6,20 @@ import { AuthModule, LogLevel, OidcConfigService } from 'angular-auth-oidc-clien
...
-      redirectUrl: window.location.origin,
-      postLogoutRedirectUri: window.location.origin,
-      scope: 'openid outofschoolapi.read',
-      logLevel: LogLevel.Debug,
-    });
+ 
+  return () =>{ 
+      oidcConfigService.withConfig({
+        clientId: 'angular',
+        stsServer: 'http://localhost:5443',
```
dmitryy90, line 13: move to environment.ts

dmitryy90 (approved):

PR #73, opened 2021-02-24
Guard+rounting+lazyloading for config/activities (provider/parent)

litvinets:


src/app/app-routing.module.ts
```diff
@@ -1,6 +1,7 @@
 import { NgModule } from '@angular/core';
 import { Routes, RouterModule } from '@angular/router';
 
+
```
dmitryy90, line 4: why?

src/app/header/header.component.ts
```diff
@@ -47,4 +48,10 @@ export class HeaderComponent implements OnInit {
   login(): void{
     this.store.dispatch(new Login());
   }
+  linksGenerating(){
+    let role = this.store.selectSnapshot<string>(UserRegistrationState.role);
+    let urlRole = (role==='organization') ? 'provider':'parent';
```
dmitryy90, line 53: you can move it to enum and use it as an associative array

src/app/shell/parent/parent.guard.ts
```diff
@@ -0,0 +1,24 @@
...
+  providedIn: 'root'
+})
+export class ParentGuard implements CanLoad {
+  constructor(public store: Store) {}
+
+  canLoad(
+    route: Route,
+    segments: UrlSegment[]): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
+      const role=this.store.selectSnapshot<string>(UserRegistrationState.role);
+    if(role==='parent'){
```
dmitryy90, line 17: return role === 'parent';

dmitryy90 (changes requested):

dmitryy90 (approved):

src/app/header/header.component.ts
litvinets: I didn't understand how or what to do;
I will do it later

Issue #99, opened 2021-03-12
User name label falls down after user is logged in


dmitryy90:
User name label falls down after user is logged in

![image](https://user-images.githubusercontent.com/1244989/110952659-26b54100-834f-11eb-9996-091bfe9797e0.png)

PR #110, opened 2021-03-15
Fix/#102 and #99 issues

litvinets:
Fix for issues 102 and 99
1. The header (template and styles )is changing depending on the routing links (main/activities/config). 
2. Merged with Yura's changes.


src/app/header/header.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="header">
+<div class="header header_main" [ngClass]="(isMainPage$ | async) ? 'header_main' : 'header_not_main'">
```
dmitryy90, line 1: please use meaningful class names

dmitryy90 (changes requested):

src/app/header/header.component.html
litvinets: class="header_main_page"
class="header_not_main_page"

Is it gonna be ok?

litvinets: @dmitryy90 
class="header_main_page"
class="header_not_main_page"

Is it gonna be ok?

dmitryy90: header > main  | header > secondary
or any descriptive class name
not_main makes no sense to me

dmitryy90 (approved):

Issue #116, opened 2021-03-23
Refactoring: remove service calls from component and use store actions instead


dmitryy90:
### Task description
Refactoring: remove service calls from component and use store actions instead

### To do 
1. - [ ] Locate all places where service calls are used
2. - [ ] Dispatch actions instead and use value from state

PR #118, opened 2021-03-25
#116 state refactoring

litvinets:
1. Two additional states were created (provider, parent)
2. Use of services were moved to the corresponding states
3. States were added to the corresponding child modules.

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -48,6 +48,7 @@ import { UserState } from './shared/store/user.state';
       //ProviderState,
      // ParentState
     ]),
+    
     NgxsReduxDevtoolsPluginModule.forRoot({
       disabled: environment.production
     }),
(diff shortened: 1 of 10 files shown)


src/app/app.module.ts
```diff
@@ -42,7 +45,10 @@ import { SharedModule } from './shared/shared.module';
       FilterState,
       MetaDataState,
       UserRegistrationState
+      //ProviderState,
```
dmitryy90, line 48: do we still need this?

src/app/shared/store/parent.actions.ts
```diff
@@ -0,0 +1,8 @@
+export class GetChildrenActivitiesList {
+  static readonly type = '[parent] gets child activities';
+  constructor() {}
+}
+export class getChildCards {
```
dmitryy90, line 5: capitalize

dmitryy90 (changes requested):

dmitryy90 (approved):

PR #130, opened 2021-04-03
Feature/#77 create a cascading select

NabokinAlexandr:


src/app/shared/components/hobby-select/hobby-select.component.html
```diff
@@ -0,0 +1,26 @@
+<div class="hobby-select">
+    <select
```
dmitryy90, line 2: we use 2space tabs

dmitryy90 (approved):

PR #155, opened 2021-04-14
Feature/updating for several tasks

litvinets:
1.Mock string files of cities and keyWords were changed to objects and functionalities of the components were changed according to the new value format 
2. Added validation for the same key word. It checks the same lower case spelling. If so - the word will not be added to the list of selected words.
3. Added validation for age and class amount input. If the length extends the required one or the input value is beyond the required one - it resets.
 (but I don't know how to check if ageFrom is not bigger than ageTo 😞. If I do checking for ageFrom - it resets because ageFrom is zero yet, but if I do checking for ageTo it resets because  if it starts with 1 (for ex: 12) - it is less than ageFrom.
dunno what to do 😟)


src/app/shell/provider/provider-activities/create-activity/create-about-form/create-about-form.component.ts
```diff
@@ -25,16 +25,44 @@ export class CreateAboutFormComponent implements OnInit {
...
+  onUploadLogo( event ):void {
     return null;
   }
+  /**
+   * This method sets validation for age and classAmount inputs and resets them if values are invalid
+   */
+  onInputValidation(): void {
+    this.AboutFormGroup.get('ageFrom').valueChanges.subscribe(val=> {
+      if(val){
+        if ( val < 1 || val > 16 ) this.AboutFormGroup.get('ageFrom').reset();
```
dmitryy90, line 45: let store magic numbers in constants with descriptive names

src/app/shell/provider/provider-activities/create-activity/create-description-form/create-description-form.component.ts
```diff
@@ -6,10 +6,9 @@ import { Select, Store } from '@ngxs/store';
 import { COMMA, ENTER } from '@angular/cdk/keycodes';
 import { MatChipInputEvent } from '@angular/material/chips';
 import { debounceTime, distinctUntilChanged, map, startWith, takeUntil} from 'rxjs/operators';
-import { KeyWordsService } from '../../../../../shared/services/key-words/key-words.service';
+import { keyWord, KeyWordsService } from '../../../../../shared/services/key-words/key-words.service';
```
dmitryy90, line 9: if it is a class it should be capitalized and moved to model file

dmitryy90 (changes requested):

src/app/shell/provider/provider-activities/create-activity/create-about-form/create-about-form.component.ts
```diff
@@ -25,16 +26,44 @@ export class CreateAboutFormComponent implements OnInit {
...
+  onUploadLogo( event ):void {
     return null;
   }
+  /**
+   * This method sets validation for age and classAmount inputs and resets them if values are invalid
+   */
+  onInputValidation(): void {
+    this.AboutFormGroup.get('ageFrom').valueChanges.subscribe(val=> {
+      if(val){
+        if ( val < Age.From || val > Age.To ) this.AboutFormGroup.get('ageFrom').reset();
```
dmitryy90, line 46: it makes no sense =) I believe minAge and maxAge are more meaningful

dmitryy90 (approved):

PR #169, opened 2021-04-20
Bug/#165 workshop naming consistent

LenkVskv:


src/app/shell/main/main.component.ts
```diff
@@ -29,12 +29,17 @@ export class MainComponent implements OnInit {
...
+  
   ngOnInit(): void {
     this.store.dispatch([new GetCategories(),
     new GetCategoriesIcons(),
     new GetPopWorkshops(),
     new ChangePage(true)
     ]);
+    
+    this.cards$.subscribe(cards =>
+      this.cards = cards
```
dmitryy90, line 41: what the reason for this change?

dmitryy90 (changes requested):

src/app/shell/main/main.component.ts
LenkVskv: I thought that the structure of all these components that contains workshop should  have the same method of realizing cards.

dmitryy90: use cards$ directly from the store through async pipe

dmitryy90 (changes requested):

dmitryy90 (approved):

PR #212, opened 2021-05-31
fix for mock API

litvinets:


src/environments/environment.ts
```diff
@@ -6,6 +6,7 @@ export const environment = {
   production: true,
   stsServer: 'http://auth.oos.dmytrominochkin.cloud',
   serverUrl: 'http://api.oos.dmytrominochkin.cloud',
+  mockUrl: 'http://localhost:4200/',
```
dmitryy90, line 9: what if somebody uses a different port or url on his local env?

src/app/shared/interceptors/http-token.interceptor.ts
```diff
@@ -30,12 +30,20 @@ export class HttpTokenInterceptor implements HttpInterceptor {
         );
     }
 
-    const token = this.oidcSecurityService.getToken();
-    const tokenTitle = (token) ? `Bearer ${token}` : null;
+    if (request.url.endsWith('.json')) {
+      return next.handle(request.clone({
+        url: environment.mockUrl + request.url,
```
dmitryy90, line 35: no need to set url for mocks

dmitryy90 (changes requested):

dmitryy90 (approved):

PR #214, opened 2021-06-01
fix fo create chidl and create teahcer

litvinets:
1. Option to delete teacher and child appears only if there more than 1 child/teacher
2. fix for  broken child form (it was duplicated)
3. fix of create-child routing

### DIFF
--- a/src/app/shell/personal-cabinet/parent/parent-create-child/child-form/child-form.component.html
+++ b/src/app/shell/personal-cabinet/parent/parent-create-child/child-form/child-form.component.html
@@ -35,7 +35,7 @@
     </mat-select>
   </mat-form-field>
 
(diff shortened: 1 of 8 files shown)


src/app/shell/personal-cabinet/parent/parent-create-child/child-form/child-form.component.html
```diff
@@ -35,9 +35,9 @@
     </mat-select>
   </mat-form-field>
 
-  <div fxLayout='row' fxLayoutAlign='center'>
+  <div fxLayout='row' fxLayoutAlign='center' *ngIf="childrenAmount>1">
```
dmitryy90, line 38: try to make a decision how to  write such expressions: 'a>1' or 'a > 1'

dmitryy90 (approved):

PR #241, opened 2021-06-06
click option

DenysIvashchenko:
added pointer and click option on the user name. Need to fix position


src/app/header/header.component.html
```diff
@@ -1,27 +1,34 @@
 <div class="header" [ngClass]="(isMainPage$ | async) ? '' : 'header_secondary'">
   <div class="header_top">
     <div class="header_logo-wrapper">
-      <a routerLink="/"><img class="logo" src="../../assets/images/Logo.svg" alt="Позашкілля"></a>
+      <a routerLink="/"
```
dmitryy90, line 4: please dont change formatting

dmitryy90 (changes requested):

Issue #238, opened 2021-06-04
Style key words accroding to the design


litvinets:
Current state:
![image](https://user-images.githubusercontent.com/27493959/120840078-6c5e6000-c572-11eb-971c-2020eaf93540.png)

Should be:
![image](https://user-images.githubusercontent.com/27493959/120839357-83508280-c571-11eb-800f-dd340a2ce50e.png)

PR #255, opened 2021-06-10
Style key words accroding to the design #238

sivxkxn:
Fixed some styles in key words.


angular.json
```diff
@@ -37,7 +36,13 @@
               "src/styles.scss",
               "./node_modules/leaflet/dist/leaflet.css"
             ],
-            "scripts": []
+            "scripts": [],
```
dmitryy90, line 39: what the reason of those chages?

package.json
```diff
@@ -15,17 +15,17 @@
   "private": true,
   "dependencies": {
     "@angular-slider/ngx-slider": "^2.0.3",
-    "@angular/animations": "~11.0.9",
+    "@angular/animations": "~12.0.4",
```
dmitryy90, line 18: do we really need to update all those libraries to have styled key words chips?

src/app/shared/styles/create-form.scss
```diff
@@ -1,50 +1,50 @@
...
   }
-  &-required{
-    color:red;
+  &-required {
+    color: red;
   }
-  &-input{
-    border: 1px solid #E3E3E3;
+  &-input {
+    border: 1px solid #e3e3e3;
```
dmitryy90, line 12: I believe you have css formatter installed, please dont use it as we disscussed

yarn.lock
```diff
@@ -12,21 +12,13 @@
     jest-preset-angular "^8.3.2"
     lodash "^4.17.15"
 
-"@angular-devkit/architect@0.1100.7":
```
dmitryy90, line 15: I can see a lot of changes that are not related to fix. Please get rid of them

dmitryy90 (changes requested):

PR #256, opened 2021-06-11
Create application

litvinets:
1. There is an opportunity to add to the application only a child (NOT YOURSELF). 
2. The layout for the case when you want to add yourself is also created.
3. Create-application component is moved to the parent folder, because only parent can do this action
4. Added API call for application creation and its handlers for the success/fails cases
5. Updated Child mock with the most relative child-model
6. Updated services (application and child services) according to the API we have
image
image

TODO:
1. fix styling and make it closer to the design view.
2. Add url parameter of the workshop (needs to be done after workshop-details is fix.) and then link the real workshopId to the application
3. confirmation modal window when you want to stop creating an application
4. Fix application in the personal cabinet, according to the new statuses from back-end
(denied -> rejected, new -> pending). And change the logic of getting information for child info-box. Instead of having a child, we need to make an API call with childId
5. Replace API call /Children/Get with /Children/GetByParentId (we can get right now only all children, not for the specific parent)
6, Updated child and application models according to the swagger


src/app/shared/components/person-card/person-card.component.html
```diff
@@ -1,33 +1,59 @@
...
-      <h4>Телефон</h4>
-      <p>{{card.phone}}</p>
-      <h4>Емейл</h4>
-      <p>{{card.email}}</p>
-    </mat-card-content>
-  </mat-card>
+
+<ng-template #ParentView>
+
+  <label class="card_label">ПІБ</label>
```
dmitryy90, line 24: when we started to use underscore in classnames? I believe we use dash in the project

src/app/shared/pipes/application-filter.pipe.ts
```diff
@@ -8,7 +8,9 @@ import { Application } from '../models/application.model';
 export class ApplicationFilterPipe implements PipeTransform {
 
   transform(array: Application[], status: string): Application[] {
-    return array.filter(card => card.status === status);
+    // return array.filter(card => card.status === status);
```
dmitryy90, line 11: if you leave commented code please use TODO: for letting other teammates know that it is needed

dmitryy90 (approved):

PR #261, opened 2021-06-12
Fix/create provider

litvinets:
I added comments to the issues below, check it

**There is a problem with create-provider guard. I don't know why, but currently, the guard returns role checking before getting user/user.role from the state. As a result, it leads to an error because of undefined.role. Nevertheless, we do have a user in the state, because we set it there right after registration. Anyway, I commented this guard because it blocked the work of create-provider component.** 

I added these fields to the create-provider 
photo_2021-06-12_15-10-16
photo_2021-06-12_15-10-20

**But I am confused about how we should keep this. I created enums according to the back-end, but for the requests, we need to set not the string, but the number. And I do not really know how to organize it in a rigth way. The same thing for workshop-type. Should I get rid of the enums at all? Or should I create not an enum but a class with its number-value and ukr-name-string?**


src/app/shell/personal-cabinet/provider/create-provider/create-provider.component.html
```diff
@@ -32,19 +32,19 @@ <h3 class="wrapper-title">РЕЄСТРАЦІЯ НОВОГО ЗАКЛАДУ</h3>
...
+          <mat-checkbox class="step-check-box" color="primary" [formControl]="AgreementFormControl">Я
             не робот
           </mat-checkbox>
         </div>
         <div fxLayout='row' fxLayoutAlign='center center' class="footer">
           <button class="btn" mat-button matStepperPrevious>Назад</button>
           <button class="btn btn-cancel" mat-raised-button><a [routerLink]="''">Скасувати</a></button>
           <button class="btn" mat-button type="submit" (click)="onSubmit()"
-            [disabled]="!isAgreed && !isNotRobot">Завершити</button>
+            [disabled]="(!isAgreed || !isNotRobot)">Завершити</button>
```
dmitryy90, line 47: I am pretty much sure that somebody is already working on this bug fix

src/app/shell/personal-cabinet/provider/provider-org-info/provider-org-info.component.html
```diff
@@ -10,25 +10,25 @@
               <mat-icon style="color:#3849F9;">edit</mat-icon>Редагувати
             </button></a>
           <h3>Форма властності</h3>
-          <p>{{provider.ownership}}</p>
+          <p> {{getOwnerShipType((provider$ | async)?.ownership)}} </p>
```
dmitryy90, line 13: it may cause a performance issue, try to create a local var from provider$ and use here

dmitryy90: <-ng-container *ngIf="provider$ | async as provider">
   <-p> {{ provider.foo}}<-/p>
   <-p> {{ provider.bar}}<-/p>
<-/ng-container>

src/app/shell/shell-routing.module.ts
```diff
@@ -34,7 +35,7 @@ const routes: Routes = [
   {
     path: 'create-provider', component: CreateProviderComponent,
     loadChildren: () => import('./personal-cabinet/provider/provider.module').then(m => m.ProviderModule),
-    canLoad: [ProviderGuard]
+    // canLoad: [CreateProviderGuard] TODO: fix guard
```
dmitryy90, line 38: lets discuss it on daily meeting

dmitryy90 (changes requested):

src/app/shell/personal-cabinet/provider/create-provider/create-info-form/create-info-form.component.ts
```diff
@@ -34,6 +41,7 @@ export class CreateInfoFormComponent implements OnInit {
 
   ngOnInit(): void {
     this.passInfoFormGroup.emit(this.InfoFormGroup);
+    console.log(this.providerTypeUkr[this.providerType.FOP])
```
dmitryy90, line 44: console?

src/app/shell/personal-cabinet/provider/provider.guard.ts
```diff
@@ -16,6 +16,7 @@ export class ProviderGuard implements CanLoad {
     route: Route,
     segments: UrlSegment[]): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
     const user = this.store.selectSnapshot<User>(RegistrationState.user);
-    return user.role === Role.provider;
+    const userRole = this.store.selectSnapshot<User>(RegistrationState.user).role;
```
dmitryy90, line 19: needs to be replaced with observable

dmitryy90 (approved):

Issue #230, opened 2021-06-04
Add checkbox for the case when actual and legal address is the same on Create Provider


litvinets:
Current state:
![image](https://user-images.githubusercontent.com/27493959/120835797-792c8500-c56d-11eb-8a7e-44846fca50f4.png)

Should be 
![image](https://user-images.githubusercontent.com/27493959/120835952-ada04100-c56d-11eb-8653-a02c75ada0e3.png)

PR #263, opened 2021-06-12
#230 checkbox for create workshop

litvinets:
1. Added functionality for the checkbox (if checked to use provider information for form fields)
2. Fixed the workshop model (there were new required fields on the back-end) and added new formControls for them
3. Added new action that can clear the state from categories. Otherwise, when we create our second workshop - we still have the same categories displayed for the user.

### DIFF
--- a/src/app/shared/models/workshop.model.ts
+++ b/src/app/shared/models/workshop.model.ts
@@ -39,7 +39,6 @@ export class Workshop {
   providerTitle?: string;
   isPerMonth?: string;
 
-
   constructor(about, description, addr, tchrs) {
     this.title = about.title;
     this.phone = about.phone;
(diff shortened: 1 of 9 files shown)


src/app/shared/models/workshop.model.ts
```diff
@@ -57,10 +60,11 @@ export class Workshop {
     this.category = description.category;
     this.categoryId = description.category.id;
     this.providerId = 1;
-
+    this.providerTitle = about.providerTitle;
```
dmitryy90, line 63: duplicated line

src/app/shared/store/meta-data.actions.ts
```diff
@@ -20,3 +20,7 @@ export class KeyWordsList {
   static readonly type = '[meta-data] Get list of key words';
   constructor(public payload: KeyWord[]) { }
 }
+export class ClearCategories {
+  static readonly type = '[meta-data] clear categories list';
```
dmitryy90, line 24: capitalize

dmitryy90 (approved):

PR #276, opened 2021-06-15
move map component from provider->create workshop  to shared folder t…

olegz7:
…o render map on search window for unregistered users


src/app/shell/result/result.component.html
```diff
@@ -34,7 +34,12 @@ <h3>Знайдено Х гуртків</h3>
   </mat-drawer>
   <mat-drawer-content [style.marginLeft]="isFiltersVisible ? '326px' : '0'">
     <div *ngIf="group.value==='map'; else ShowData">
-      <app-map></app-map>
+      <app-map>
+        <!-- <div style="height: 500px;"
```
dmitryy90, line 38: do we need this?

dmitryy90 (changes requested):

src/app/shell/result/result.component.html
olegz7: of coarse not ))
fixed, pushed

dmitryy90 (approved):

PR #324, opened 2021-07-04
General fix of the functionality for edit User Mode

litvinets:
Cause we update the personal info both for Parent and Provider, the update flow for replaced with user-service.

(previously we used Parent-Update API for this, but this process is the same for both roles).

The general functionality was redone throw the state-actions.


src/app/shared/services/user/user.service.ts
```diff
@@ -15,4 +15,12 @@ export class UserService {
   getUserById(id): Observable<User> {
     return this.http.get<User>(`/User/GetUserById/${id}`);
   }
+
+  /**
+  * This method update Provider
+  * @param Provider
+  */
+  updateUser(user: User): Observable<Object> {
+  return this.http.put('/User/Update', user);
```
dmitryy90, line 24: indentations

dmitryy90 (approved):

PR #330, opened 2021-07-05
fix edit user mode

litvinets:
Edit User works:))

### DIFF
--- a/src/app/shared/models/user.model.ts
+++ b/src/app/shared/models/user.model.ts
@@ -15,7 +15,6 @@ export class User {
     this.lastName = info.lastName;
     this.middleName = info.middleName;
     this.phoneNumber = info.phoneNumber;
-    this.email = info.email;
     this.id = id;
   }
 }
--- a/src/app/shared/models/user.model.ts
+++ b/src/app/shared/models/user.model.ts
@@ -1,4 +1,4 @@
-export interface User {
+export class User {
   isRegistered: boolean;
   lastName?: string;
   middleName?: string;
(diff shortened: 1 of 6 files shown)


src/app/shared/models/user.model.ts
```diff
@@ -8,4 +8,13 @@ export interface User {
   email?: string;
   phoneNumber?: string;
   role: string;
-}
\ No newline at end of file
+
+  constructor(info, id?: string) {
+    this.id = info.id;
```
dmitryy90, line 13: you put id twice
is there any reason to pass id into constructor?

dmitryy90 (changes requested):

src/app/shared/models/user.model.ts
litvinets: yeah, otherwise we cannot update the user. it is required by backend. that's why I deleted the interface and created class

litvinets: and I don't see the reason to add the id to the Reactive Form group. ofc I can pass this body directly to the service without Object creation, but I've decided to follow the standard flow as for other items (workshop, provider)

src/app/shared/models/user.model.ts
```diff
@@ -8,4 +8,12 @@ export interface User {
...
   role: string;
-}
\ No newline at end of file
+
+  constructor(info, id?: string) {
+    this.firstName = info.firstName;
+    this.lastName = info.lastName;
+    this.middleName = info.middleName;
+    this.phoneNumber = info.phoneNumber;
+    this.id = id;
```
dmitryy90, line 17: if id is not passed it could be set as undefined, so it needs some condition here

dmitryy90 (changes requested):

src/app/shared/models/user.model.ts
litvinets: then I have changed it to the required because we don't create an initial user, we always get it from the back-end with id by default. So, when we edit the user, it for sure has its id.

dmitryy90 (approved):

PR #337, opened 2021-07-08
Get cities

litvinets:
If you tap more than 3 characters, the action GetCities is dispatched
If you tap beleberdu, then to cities state is set "No such city"
If the input is empty, we clear the state, so we won't have there "No such City"

**NOTE!**

We dispatch ClearCity all the time if we have the !value.length>2 
image
If we tap 1 symbol, then another one, or delete one character and again. 
Is it ok??


src/app/shared/components/city-autocomplete/city-autocomplete.component.ts
```diff
@@ -1,84 +1,48 @@
...
-      if (value) {
-        this.store.dispatch(new CityList(this._filter(value)));
-      } else {
-        this.store.dispatch(new CityList([]));
-      }
-      ;
-    });
-    this.address && this.address.valueChanges.subscribe(address => this.cityControl.setValue(address.city));
+      ).subscribe(value =>
+        (value.length > 2) ? this.store.dispatch(new GetCities(value)) : this.store.dispatch(new ClearCities()));
```
dmitryy90, line 44: move check to filter operator;
why do we need to clear cities?

src/app/shared/store/meta-data.state.ts
```diff
@@ -105,5 +107,18 @@ export class MetaDataState {
     patchState({ departments: undefined });
     patchState({ classes: undefined });
   }
+  @Action(GetCities)
+  getCities({ patchState }: StateContext<MetaDataStateModel>, { payload }: GetCities) {
+    return this.cityService
+      .getCities(payload)
+      .pipe(
+        tap((cities: City[]) => cities ? patchState({ cities: cities, isCity: true }) : patchState({ cities: [{ name: 'Такого міста немає' } as City], isCity: false })
```
dmitryy90, line 115: you can move ternary inside function

dmitryy90 (changes requested):

src/app/shared/components/city-autocomplete/city-autocomplete.component.ts
litvinets: @dmitryy90  because when I for example tapped something unexisting, then I need to display no such city. I push this thing to state. But then if I clear the input again and won't make new GetCities yet, I will still see "No such City", because we have it from teh state

litvinets: and we do not have the filtration on the front-end, it's on the back-end. 
If I tap 3-5 characters I will have no more than 8-12 cities, usually, it's 8. SO I don't see the point to make the filtration again.

litvinets: that's why I have deleted filtration at all. Cause I thought we don't need to make filtration for 12 cities again.
Should I revet it?

dmitryy90: I mean put length check in pipe(filter()) for avoiding triggering of subscription

dmitryy90: lets talk on tomorrow's meeting about clear cities action

dmitryy90 (approved):

PR #355, opened 2021-07-13
Application update

litvinets:


src/app/shared/registration/registration.module.ts
```diff
@@ -13,7 +13,7 @@ export function configureAuth(oidcConfigService: OidcConfigService): any {
     oidcConfigService.withConfig({
       useRefreshToken: true,
       silentRenew: true,
-      silentRenewTimeoutInSeconds: 10,
+      silentRenewTimeoutInSeconds: 2000,
```
dmitryy90, line 16: 2000 seconds ~ 33 minutes, did you say that we have 5 min token?

dmitryy90 (approved):

PR #374, opened 2021-07-19
images are replaced with mat-icons

litvinets:
Among all the icons, I found only the "approved" status, but not the others. Should we ask to replace these icons with those that we can have from mat-icons, or should we use images for displaying these statuses? 
https://www.angularjswiki.com/angular/angular-material-icons-list-mat-icon-list/
image

### DIFF
--- a/src/app/shared/components/workshop-card/workshop-card.component.scss
+++ b/src/app/shared/components/workshop-card/workshop-card.component.scss
@@ -9,7 +9,7 @@
 .action-icon {
   color: #fff;
 }
-.status{
+.parent-status{
   .approved{
     color:#199953
   }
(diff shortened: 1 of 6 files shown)


src/app/shared/components/workshop-card/workshop-card.component.scss
```diff
@@ -1,5 +1,23 @@
...
-
-.mat-icon {
+.icon{
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  align-items: center;
+  font-weight: 500;
+}
+.action {
```
dmitryy90, line 9: use namespace

dmitryy90 (approved):

PR #393, opened 2021-07-27
Rating final update

litvinets:
If the role is provider, so rating button so not displayed.
If the user is not logged in, so on click it redirects to login page.

Added users' names and their rate as stars. At the new rate, the actual rate of the workshop is updated.
image

### DIFF
--- a/src/app/shared/store/user.state.ts
+++ b/src/app/shared/store/user.state.ts
@@ -448,7 +448,7 @@ export class UserState {
   @Action(OnCreateRatingFail)
   onCreateRatingFail({ dispatch }: StateContext<UserStateModel>, { payload }: OnCreateRatingFail): void {
     throwError(payload);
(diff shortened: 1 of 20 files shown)


src/app/shared/store/user.state.ts
```diff
@@ -451,35 +448,12 @@ export class UserState {
   @Action(OnCreateRatingFail)
   onCreateRatingFail({ dispatch }: StateContext<UserStateModel>, { payload }: OnCreateRatingFail): void {
     throwError(payload);
-    dispatch(new ShowMessageBar({ message: 'На жаль виникла помилка', type: 'error' }));
+    dispatch(new ShowMessageBar({ message: 'Ви вже оцінили цей гурток', type: 'warningYellow' }));
```
dmitryy90, line 451: I believe we should not allow accessing control if user already voted

dmitryy90 (changes requested):

dmitryy90 (approved):

PR #459, opened 2021-09-08
Language change dropdown bug#438/zubkov

olegz7:


src/app/app.component.ts
```diff
@@ -1,14 +1,29 @@
...
   styleUrls: ['./app.component.scss']
 })
 
-export class AppComponent {
-  title = 'out-of-school';
-  constructor () {}
+export class AppComponent implements OnInit{
+  constructor() { }
+
+  MobileView: boolean = false;
```
dmitryy90, line 12: this needs to be removed from header component and be passed as input parameter from app component into header conponent

dmitryy90 (changes requested):

src/app/app.component.ts
olegz7: fixed, did same thing for sidenav component

dmitryy90: i dont see any changes

dmitryy90 (changes requested):

src/app/shared/components/sidenav/sidenav.component.ts
```diff
@@ -59,6 +59,7 @@ export class SidenavComponent implements OnInit, OnDestroy{
   }
 
   ngOnInit(): void {
+    this.selectedLanguage = localStorage.getItem('ui-culture') || 'uk';
     this.user$.subscribe(user => this.user = user);
     this.isWindowMobile(window);
```
dmitryy90, line 64: you can remove handlers if you pass movileView param

dmitryy90 (changes requested):

dmitryy90 (approved):

PR #464, opened 2021-09-15
No results wrapper

litvinets:
1. Shared folder is a bit refactored (module is moved to the module folder)
2. AppState is cleaned from unnecessary services, values, imports. AppState Loading is removed because it is not used anymore
3. isLoadingProviderCabinet is renamed to isLoadingCabinet
4. Added component for no-result and added to the templates
5. removed unnecessary GetAllWorkshops because it is not used anymore


src/app/shared/components/no-result-card/no-result-card.component.ts
```diff
@@ -0,0 +1,17 @@
+import { Component, Input, OnInit } from '@angular/core';
+
+@Component({
+  selector: 'app-no-result-card',
+  templateUrl: './no-result-card.component.html',
+  styleUrls: ['./no-result-card.component.scss']
+})
+export class NoResultCardComponent implements OnInit {
```
dmitryy90, line 8: JFYI you may use inline template if the component has no logic and styles and is pretty short

dmitryy90 (approved):

PR #483, opened 2021-09-22
fix

litvinets:
1. If it is only one page of result -> don't show pagination
2. view fix of the result page
3. display the amount of found workshops


src/app/shared/components/paginator/paginator.component.html
```diff
@@ -1,4 +1,4 @@
-<div fxLayout='row' fxLayoutAlign='center center'>
+<div *ngif="totalPageAmount > 1" fxLayout='row' fxLayoutAlign='center center'>
```
dmitryy90, line 1: ngIf

dmitryy90 (approved):

src/app/shared/components/paginator/paginator.component.html
litvinets: glaz almaz!

PR #484, opened 2021-09-23
City confirmation

litvinets:


src/app/shared/components/city-autocomplete/city-autocomplete.component.ts
```diff
@@ -54,9 +54,15 @@ export class CityAutocompleteComponent implements OnInit {
...
         filter(value => value?.length > 2)
       ).subscribe(value => this.store.dispatch(new GetCities(value)));
+    this.InitialCity && this.setInitialCity();
 
-    this.InitialCity && this.setInitialAcity();
   }
+
+
+  ngOnChanges(): void {
+    this.InitialCity && this.setInitialCity();
```
dmitryy90, line 63: this won't work

I can see that it will be creating multiple subscriptions for GetCities every time user opens main page wich may cause performance issue;

also, I assume it may prevent users from selecting city manually

dmitryy90 (changes requested):

PR #509, opened 2021-10-04
refactor support about

romandivonchuk:


src/app/shell/info/about/about.component.scss
```diff
@@ -0,0 +1,99 @@
+// .about {
```
dmitryy90, line 1: do we still need commented code?

dmitryy90 (changes requested):

src/app/shell/info/about/about.component.scss
romandivonchuk: No, my mistake

dmitryy90 (approved):

PR #521, opened 2021-10-08
Bug fix/#435 favorite workshop bugs

DenysIvashchenko:


src/app/shared/components/workshop-card/workshop-card.component.ts
```diff
@@ -92,8 +103,25 @@ export class WorkshopCardComponent implements OnInit, OnDestroy {
...
     this.destroy$.next(true);
     this.destroy$.unsubscribe();
   }
 }
+
+@Component({
+  selector: 'app-workshop-dialog',
+  template: `
+    <div mat-dialog-content fxLayoutAlign="center" class="dialog-title">
+      <p>Для того щоб додати в улюблені будь ласка зареєструйтесь на порталі. Дякуемо!</p>
```
litvinets, line 120: fix spelling "Дякуємо"

src/app/shell/main/main.component.ts
```diff
@@ -34,31 +41,49 @@ export class MainComponent implements OnInit {
...
+    if(role === Role.parent) {
+      combineLatest([this.city$, this.favoriteWorkshops$])
+      .pipe(
+        filter((result) => {
+          if(!!result[0] && !!result[1]?.length) {
+            return true;
+          }
+          else if(result[1] === null) {
+            return true;
+          }
```
litvinets, line 60: I guess you can simplify it to this
filter((result) => (!!result[0] && !!result[1]?.length) || (result[1] === null))

@dmitryy90  verify it please:)

src/app/shell/main/main.component.ts
```diff
@@ -34,31 +41,49 @@ export class MainComponent implements OnInit {
...
+      .subscribe(parent => this.parent = parent);
 
-    this.isParent$
+    this.role$
       .pipe(
+        filter(role => !!role),
         takeUntil(this.destroy$))
-      .subscribe(parent => this.parent = parent);
+      .subscribe(role => this.getTopWorkshops(role))  
   }
```
litvinets, line 87: I truly believe you don't need it anymore. It is not used in the ts. The "parent" is passed to the workshop card and as a boolean (that is not correct). So we do not really need the subscribtion to the parent. The parent input should be removed and the logic for the view should be replaces with the role-state. 
image

src/app/shell/shell.component.ts
```diff
@@ -28,20 +28,21 @@ export class ShellComponent implements OnInit, OnDestroy {
...
     });
 
     this.isParent$
-      .pipe(takeUntil(this.destroy$))
-      .subscribe((parent) => {
-      !!parent && this.store.dispatch([
+      .pipe(takeUntil(this.destroy$),
+        filter(parent => !!parent))
+      .subscribe(() => {
+       this.store.dispatch([
```
litvinets, line 45: I also see the you do not need the actual instance of parent, you just need to coordinate the action depending on the role. In this case this actions should be dispatched if the user as parent is received and we have his/her profile. I suggest to replace this subscription also to the role state subscription.

src/app/shared/store/registration.state.ts
```diff
@@ -111,20 +120,24 @@ export class RegistrationState {
...
           ));
     } else {
       return this.providerService
         .getProfile()
         .pipe(
           tap(
-            (provider: Provider) => patchState({ provider: provider })
+            (provider: Provider) => {
+              patchState({ provider: provider });
+          }
```
litvinets, line 140: You duplicate teh same action twice. You patch the state when you got user and then is doing it again in if/else.
I suggest to move  patchState({ role: state.user.role}); after if/else and use it only once. Or remove 123 line.

src/app/shared/components/workshop-card/workshop-card.component.ts
```diff
@@ -58,6 +65,10 @@ export class WorkshopCardComponent implements OnInit, OnDestroy {
         this.favoriteWorkshopId = this.favoriteWorkshops?.find(item => item.workshopId === this.workshop?.workshopId);
       });
     this.isFavorite = !!this.favoriteWorkshopId;
+
+    this.provider$
+      .pipe(takeUntil(this.destroy$))
+      .subscribe((provider)=> this.isProvider = !!provider);
```
litvinets, line 71: you do not need it. you don't use the inctance of provider (you don't use its id/name/ etc). you just need the role for detecting the view type. so use please the useRole input

src/app/shell/main/main.component.ts
dmitryy90: not sure what you are trying to achieve with this condition. if workshops === null, the first condition must throw an error.

to avoid using indexes you can use destructing assignment:
const city, workshops;
[city, workshops] = result;

dmitryy90 (changes requested):

PR #539, opened 2021-10-13
Categories icons/#522/makeev

gmakeev:
The categories icons have been added to the project, enum and match the appropriate category id with the icon have been created.


litvinets:
Replace please this icon with mat-icon 
image
Apply please this one instead
<i class="fas fa-filter"></i>

PR #551, opened 2021-10-19
review

romandivonchuk:


.vscode/launch.json
```diff
@@ -0,0 +1,25 @@
+{
```
dmitryy90, line 1: please remove .vscode folder from changes or add to gitignore

src/environments/environment.ts
```diff
@@ -3,7 +3,7 @@
 
 // The list of file replacements can be found in `angular.json`.
 export const environment = {
-  production: true,
+  production: false,
```
dmitryy90, line 6: why do we need this change?

dmitryy90 (changes requested):

PR #594, opened 2021-11-01
Task 583. The style of the tabs has been fixed after doing the 'Submi…

gmakeev:
…t Application'


src/app/shell/personal-cabinet/parent/create-application/create-application.component.scss
```diff
@@ -57,7 +57,7 @@ button[disabled]{
   opacity: 1;
   background-color: transparent;
 }
-::ng-deep .mat-tab-header {
+:host ::ng-deep .mat-tab-header {
```
dmitryy90, line 60: I can see other risky places in the file, it would be better to wrap all styles

dmitryy90 (changes requested):

dmitryy90 (approved):

PR #599, opened 2021-11-02
Task 598. 'Workshop Leave' link of workshop card layout when is visib…

gmakeev:
…le has been fixed.

### DIFF
--- a/src/app/shared/components/workshop-card/workshop-card.component.scss
+++ b/src/app/shared/components/workshop-card/workshop-card.component.scss
@@ -71,9 +71,15 @@
   border-radius: 20px;
 }
 
+.card {
+  &_link {
+    margin-right: 0;
+  }
+}
+
 @media (max-width: 750px) {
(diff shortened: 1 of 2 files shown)


src/app/shared/components/workshop-card/workshop-card.component.scss
```diff
@@ -71,9 +71,15 @@
   border-radius: 20px;
 }
 
+.card {
+  &_link {
+    margin-right: 0;
```
dmitryy90, line 76: not sure how this is going to fix issue.
we already have margin-right specified in cards.css, so I believe changes should be done there.
anyway this does not solve the issue since it just breaks status and "leave workshop" link into 2 lines which causes another problem with the height of the card. as the result each card has different height

please take a look at the problem one more time and try to find another solution

dmitryy90 (changes requested):

dmitryy90 (approved):

PR #677, opened 2021-11-24
Disable keywords input when more than 5

tiana25:


litvinets (approved):

litvinets (commented):
remove please imports from ts file that are not used

dmitryy90 (approved):

PR #678, opened 2021-11-25
[Edit workshop] User sporadically is not able to switch between tabs on workshop edit #666

romandivonchuk:


src/environments/environment.ts
```diff
@@ -3,7 +3,7 @@
 
 // The list of file replacements can be found in `angular.json`.
 export const environment = {
-  production: true,
+  production: false,
   stsServer: 'https://auth.oos.dmytrominochkin.cloud',
```
litvinets, line 7: these changes are applied in another PR

dmitryy90 (approved):

PR #684, opened 2021-11-25
disabled dropdown if no workshops

tiana25:


src/app/shared/components/workshop-checkbox-dropdown/workshop-checkbox-dropdown.component.scss
```diff
@@ -19,6 +19,10 @@
   padding: 1rem;
 }
 
+.label-disabled {
+  opacity: 0.3;
```
dmitryy90, line 23: could you please check if we use such styles for all disabled fields to know if we can move it out and reuse

dmitryy90 (changes requested):

src/styles.scss
```diff
@@ -142,6 +142,10 @@ $config: mat-typography-config(
   white-space: normal !important;
 }
 
+.disabledField {
```
dmitryy90, line 145: we dont use camel case in class names
use disabled-field instead

dmitryy90 (approved):

PR #688, opened 2021-11-26
Fix the width of Number phone code, it should be the same as others

SkyLightSK:


src/app/shared/constants/constants.ts
```diff
@@ -17,7 +17,7 @@ export class Constants {
   static readonly MAX_TIME = '23:59';
   static readonly MAX_DESCRIPTION_LENGTH = 500;
   static readonly MAX_TEACHER_DESCRIPTION_LENGTH = 300;
-  static readonly PHONE_LENGTH = 10;
```
dmitryy90, line 20: I ve already seen same chages in another PR could you please validate if they are still needed?

dmitryy90 (changes requested):

Issue #753, opened 2022-01-13
[Providers cabinet - application] Workshop name overlaps


dmitryy90:

LeraKornienko:
done

PR #778, opened 2022-01-18
#753 Workshop name overlaps

LeraKornienko:
Workshop name overlaps


litvinets:
you don't need just to cut the title, you should show the full title on hover, as we do for all other cases. Check how it's done for workshopcard

litvinets:
image

litvinets (approved):

Issue #777, opened 2022-01-18
When you go to another workshop in workshop-details, images of the workshop does not changes. 


litvinets:
![image](https://user-images.githubusercontent.com/27493959/149936115-52c39cdf-97cd-443d-9e6f-4abb227ad402.png)

PR #785, opened 2022-01-20
Fix/upload relevant images#777

kokole110:
https://github.com/ita-social-projects/OoS-Frontend/issues/777

Add carousel component to shared module
Set transform style property of dynamically created div (that wraps all loaded images) to 'none' when onChange hook of image-carousel component fires to make newly loaded image(-s) sit in right place. Also set dotsArr property of Carousel component to fit actual image arr length

### DIFF
--- a/src/app/shared/components/image-carousel/image-carousel.component.ts
+++ b/src/app/shared/components/image-carousel/image-carousel.component.ts
@@ -17,9 +17,8 @@ export class ImageCarouselComponent implements OnInit, OnChanges {
 
   ngOnChanges() {
     if (this.myCarousel) {
-      this.myCarousel.utils.carouselProperties.cellsElement.style.transform = 'none'
-      this.myCarousel.dotsArr = [];
-      this.myCarousel.dotsArr = Array(this.images.length).fill(1)
+      this.myCarousel.utils.carouselProperties.cellsElement.style.transform = 'none';
+      this.myCarousel.dotsArr = Array(this.images.length).fill(1);
     }
   }
 
--- a/src/app/shared/components/image-carousel/image-carousel.component.ts
+++ b/src/app/shared/components/image-carousel/image-carousel.component.ts
@@ -1,0 +1,26 @@
+import { Component, OnInit, Input, ViewChild, OnChanges } from '@angular/core';
+import { imgPath } from 'src/app/shared/models/carousel.model';
+
+@Component({
+  selector: 'app-image-carousel',
+  templateUrl: './image-carousel.component.html',
+  styleUrls: ['./image-carousel.component.scss']
+})
+export class ImageCarouselComponent implements OnInit, OnChanges {
+
+  @Input() images:imgPath[] = [];
+  @ViewChild('myCarousel', {static: false}) myCarousel;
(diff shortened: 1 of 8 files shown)


src/app/shared/components/image-carousel/image-carousel.component.ts
```diff
@@ -0,0 +1,26 @@
...
+  @Input() images:imgPath[] = [];
+  @ViewChild('myCarousel', {static: false}) myCarousel;
+  constructor() { }
+
+  ngOnInit(): void {  
+  }
+
+  ngOnChanges() {
+    if (this.myCarousel) {
+      this.myCarousel.utils.carouselProperties.cellsElement.style.transform = 'none'
```
dmitryy90, line 20: ;

dmitryy90: not clear why do we need change styles in ts

src/app/shell/workshop-details/workshop-page/workshop-page.component.ts
```diff
@@ -7,41 +7,51 @@ import { Provider } from 'src/app/shared/models/provider.model';
...
+      this.getWorkshopImages();
+    })
   }
 
   private getWorkshopImages(): void {
     if (this.workshop?.imageIds.length) {
       this.images = this.workshop.imageIds.map((imgId) => { return { path: this.authServer + this.imgUrl + imgId } })
     } else {
+      this.images = [];
       this.images.push({ path: 'assets/images/groupimages/workshop-img.png' })
```
dmitryy90, line 55: semicolon;

 this.images = [{ path: 'assets/images/groupimages/workshop-img.png' }];

kokole110: done

src/app/shared/components/image-carousel/image-carousel.component.ts
kokole110: > not clear why do we need change styles in ts

When user loads workshop-details page for the first time the tree of dom elements inside Carousel component get created. It looks like this:

Dots and arrows divs get created if we have more than one image in carousel.

carousel bug

Case 1: initially workshop with 1 image loads, carousel with one image get created, **Carousel.dotsArr = [1]**, div.carousel-cells style property has default values ----------->
user navigates to another workshop with 3 images: images load correctly, arrows get created, but **Carousel.dotsArr = [1]** and user sees only **one** dot instead of three; when he clicks arrow to slide to the next image,  **div.carousel-cells gets this style property**: 
	transform: translateX(-788px); //some pixels here 
	transition: transform 200ms ease-out 0s;
------------------>
then user goes to another workshop with one image: 
**Carousel.dotsArr = [1]**,  
**div.carousel-cells.style:**
        transform: translateX(-788px); //some pixels here 
	transition: transform 200ms ease-out 0s;

Because this style properties stays the same, user doesn’t see image in carousel because it lays (--788px) out of the screen
So carousel stays broken until user reload the page

dmitryy90 (approved):

litvinets (approved):

PR #786, opened 2022-01-20
Release 2 feature on for debug mode and off for promode

Olya-web7:
release-2-feature-on-for-debug-mode-and-off-for-pro-mode


src/app/shared/store/meta-data.state.ts
```diff
@@ -236,8 +237,11 @@ export class MetaDataState {
   getFeaturesList({ patchState }: StateContext<MetaDataStateModel>, { }: GetFeaturesList): Observable<FeaturesList> {
     return this.featureManagementService
       .getFeaturesList()
+
       .pipe(
-        tap((featuresList: FeaturesList) => patchState({ featuresList: { release1: true, release2: true, release3: false } })
+        tap((featuresList: FeaturesList) => 
+        environment.production ? environment.production :
+        patchState({ featuresList: { release1: true, release2: true, release3: false } })
         ))
```
litvinets, line 245: I belive it should be 
```suggestion
        tap((featuresList: FeaturesList) => 
       
        patchState( environment.production ? { featuresList: featuresList} : { featuresList: { release1: true, release2: true, release3: false } })
        ))
```

litvinets (changes requested):

src/app/shared/store/meta-data.state.ts
litvinets: I see if prod true then you do nothing?:) Just checking again true prod, and if false, then you patch state. You should patch state in both cases but for prod-false patch it with values { release1: true, release2: true, release3: false }, and for prod-true with false from the response { featuresList: featuresList}

litvinets:
Assign please @dmitryy90  also as a reviewer each time:)

litvinets (approved):

Issue #799, opened 2022-01-25
[Create-Workshop] Toggle button


litvinets:
The toggle button when is toggled should have a blue background 
![image](https://user-images.githubusercontent.com/27493959/150981024-d205b67c-73c4-40fc-8fea-c6d6edb68fcc.png)
![image](https://user-images.githubusercontent.com/27493959/150981209-aa9bdf2a-6901-40cc-808c-8fa7e1877a4c.png)

There is a bug, when you first go to the result page and then to create-workshop, this toggle button becomes too wide.
![image](https://user-images.githubusercontent.com/27493959/150981159-c4bb2538-f567-47ef-a921-b03cdb3427d4.png)

PR #808, opened 2022-01-26
fix bgcolor and size of toggled buttons 

kokole110:
fix bgcolor of toggled buttons and size of 'price per lesson-month' button in create workshop
issue: https://github.com/ita-social-projects/OoS-Frontend/issues/799

I found out that commenting out some css properties works well for this task, but that properties are closely related to different toggle buttons throughout the application, so I tried to check every single page/form creation and haven't found any other unwanted changes in buttons appearance but to be safe I haven't deleted them, just commented out for now

### DIFF
--- a/src/app/shared/components/stars/stars.component.scss
+++ b/src/app/shared/components/stars/stars.component.scss
@@ -18,8 +18,7 @@
   color: #3849F9;
 }
 ::ng-deep .mat-button-toggle-checked {
-  //background-color: white!important;
-   .star-icon{
+  .star-icon{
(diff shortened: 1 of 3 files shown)


litvinets:
don't think that these styles we will need because all toggle buttons should have the same design. So, I think you can delete it

litvinets (changes requested):
don't think that these styles we will need because all toggle buttons should have the same design. So, I think you can delete it

dmitryy90 (approved):

litvinets (approved):

PR #823, opened 2022-01-28
When the user creates more than two applications for the same child, …

Olya-web7:
When the user creates more than 10 applications for the same child, then show a warning message that you are not allowed it

### DIFF
--- a/src/app/shared/store/user.state.ts
+++ b/src/app/shared/store/user.state.ts
@@ -330,8 +330,8 @@ export class UserState {
 
   @Action(OnCreateApplicationFail)
   onCreateApplicationFail({ dispatch }: StateContext<UserStateModel>, { payload }: OnCreateApplicationFail): void {
-    throwError(payload);
-    dispatch(new ShowMessageBar({ message: 'На жаль виникла помилка', type: 'error' }));
+    throwError(payload);    
+    dispatch(new ShowMessageBar({ message: 'Ліміт заяв неревищено (>10), повторіть, будь ласка, спробу пізніше', type: 'error' }));
   }
 
   @Action(OnCreateApplicationSuccess)


src/app/shared/store/user.state.ts
```diff
@@ -330,8 +331,11 @@ export class UserState {
 
   @Action(OnCreateApplicationFail)
   onCreateApplicationFail({ dispatch }: StateContext<UserStateModel>, { payload }: OnCreateApplicationFail): void {
-    throwError(payload);
-    dispatch(new ShowMessageBar({ message: 'На жаль виникла помилка', type: 'error' }));
+    throwError(payload);        
+    dispatch(new ShowMessageBar({ message: payload.error.status === 429 
+      ? `Ліміт заяв перевищено, повторіть спробу через 6 днів`
```
dmitryy90, line 336: display actual time as discussed

dmitryy90 (changes requested):

src/app/shared/store/user.state.ts
Olya-web7: Need some tip how to get "Retry-After" in user.state.ts

dmitryy90: https://stackoverflow.com/questions/36098913/convert-seconds-to-days-hours-minutes-and-seconds

Olya-web7: should I get the Retry-After from payload or add  some another parameter?

Olya-web7: payload.headers.get('retry-after') - is it correct way to get seconds for next attempt? as I get now null and headers in payload looks like: 
headers: Map(2)
[[Entries]]
0: {"content-length" => Array(1)}
1: {"content-type" => Array(1)}
size: 2

Olya-web7:
the message should look like:
MoE  Dev Team (Copy) – Figma - Google Chrome

src/app/shared/services/applications/application.service.ts
```diff
@@ -54,9 +54,48 @@ export class ApplicationService {
    * @param Workshop Workshop
    */
   createApplication(application: Application): Observable<object> {
-    return this.http.post('/api/v1/Application/Create', application);
+    return this.http.post('/api/v1/Application/Create', application, {observe: 'response'});
   }
 
+  secondsToDh(seconds: number): string {
```
dmitryy90, line 60: it may be moved to utils service

dmitryy90 (approved):

PR #853, opened 2022-02-01
About portal

tiana25:


src/app/shared/models/aboutPortal.model.ts
```diff
@@ -1,13 +1,24 @@
...
+}
+export class AboutPortal {
+  id?: string;
+  title?: string;
+  aboutPortalItems: AboutPortalItem [];
+
+  constructor(title: string, portalItem: AboutPortalItem[], id?) {
+    this.id = id;
+    this.title = title;
+    this.aboutPortalItems = portalItem;
```
dmitryy90, line 22: use plural in namings if it is an array

dmitryy90 (approved):

Issue #848, opened 2022-01-31
Minor inconsistencies in "ВИДАЛИТИ ГУРТОК?" pop-up message according to mockup


hanna-lenko:
**Environment:** Win10 v21H2 64-bit, Google Chrome v97.0.4692.99.
**Reproducible:** always
**Build found:** Frontend: gcp-0.0.50

**Preconditions**
1. Log into OoS (https://oos.dmytrominochkin.cloud/#/) as Service Provider ('email'="worms.airsoft@gmail.com", 'Password'="1qaz@WSX")
2. There is at least 1 registered workshop by Service Provider (@Гурток1)

**Steps to reproduce**
1. Go to ‘Мої гуртки’ tab of the personal cabinet 
2. Click on ‘Trash’ button in the right top corner of the workshop card from preconditions (@Гурток1)

**Actual result**
The following pop-up message appears, inconsistencies according to mockup are highlighted below
![image](https://user-images.githubusercontent.com/97878413/151788736-1e2c8555-f343-48d2-ad24-7fae6f50fbb7.png)

**Expected result**
The pop-up message appears, text corresponds to design
https://www.figma.com/file/YRYP4zAAchOyYMgdoeP782/MoE-%7C-Dev-Team?node-id=4329%3A3483
![image](https://user-images.githubusercontent.com/97878413/151788768-55928b30-0c8d-4788-8185-f6fbe0176dc9.png)

**User story and test case links**
User story #[174](https://github.com/ita-social-projects/OoS-Backend/issues/174) (Page of a section/group/class in Cabinet SP) 
[Test case] [OS-203](https://jira.softserve.academy/browse/OS-203)

**Labels to be added**
"Bug", "UI", Priority: very low, Severity: trivial, Type ("UI").

kokole110:
@hanna-lenko fixed

hanna-lenko:
@kokole110  Verified, closed

PR #854, opened 2022-02-01
fix modal window for delete workshop confirmation

kokole110:
Issue: https://github.com/ita-social-projects/OoS-Frontend/issues/848


kokole110:
@litvinets @dmitryy90  should I remove quotation mark as well (just notice they are absent on design)? If yes, remove them only for ModalConfirmationType.delete?

dmitryy90 (approved):

PR #997, opened 2022-03-14
notifications webSocket

litvinets:


src/app/shared/services/signalR/signal-r.service.ts
```diff
@@ -0,0 +1,39 @@
...
+    };
+
+    this.hubConnection = new signalR.HubConnectionBuilder()
+      .configureLogging(signalR.LogLevel.Information)
+      .withUrl(this.url, options)
+      .build();
+
+    this.hubConnection
+      .start()
+      .then(() => console.log('Connection started'))
```
dmitryy90, line 32: FYI: you can use console error, info, warn to make it more clear

dmitryy90 (approved):

PR #1001, opened 2022-03-16
Fix cover size

litvinets:
Fix image cover 
image


src/app/shared/components/child-card/child-card.component.scss
```diff
@@ -1,5 +1,7 @@
 @import "src/app/shared/styles/cards.scss";
-
+.card{
```
dmitryy90, line 2: space

dmitryy90 (approved):

Issue #979, opened 2022-03-10
[UI] No indentation in phone number


OksanaGuz:
**Environment:** macOS Big Sur 11.5.1; Chrome version 99.0.4844.51.
**Reproducible:** always.
**Build found:**
[gcp-0.0.116](https://github.com/ita-social-projects/OoS-Backend/releases/tag/gcp-0.0.116)
[gcp-0.0.55](https://github.com/ita-social-projects/OoS-Frontend/releases/tag/gcp-0.0.55)

**Preconditions**

1. Follow link https://oos.dmytrominochkin.cloud/#/
3. Log in as SP - Email: kamin@gmail.com; Password: Qwerty1!
4. Go to "ОСОБИСТА ІНФОРМАЦІЯ" page by clicking in the top right corner drop-down list

**Steps to reproduce**
1. Take a took to the phone number

**Actual result**
There is no indentation
<img width="393" alt="Снимок экрана 2022-03-10 в 17 05 34" src="https://user-images.githubusercontent.com/97882565/157690399-39c48ecf-b7be-455b-a6f1-9de82d3036f0.png">

**Expected result**
 Indentation should be implemented.
<img width="259" alt="Снимок экрана 2022-03-10 в 17 08 19" src="https://user-images.githubusercontent.com/97882565/157690988-79bc4cdb-8b3a-40f5-8691-77af67853564.png">

**User story and test case links**
US https://github.com/ita-social-projects/OoS-Backend/issues/173
TC https://jira.softserve.academy/browse/OS-358
**Labels to be added**
"Bug", Priority ("pri: middle "), Severity ("severity:trivial"), Type UI.

PR #1047, opened 2022-04-08
[UI] No indentation in phone number #979

Olya-web7:
[UI] No indentation in phone number #979


dmitryy90 (approved):

perepichai:
I created and use this pipe for the cases like this 097-382-82-79 :) Maybe it will be better to use new one

Issue #905, opened 2022-02-11
['Редагування закладу' page] 'Далі' button stays enabled when mandatory field is empty


AnnaDidiaieva:
**Environment:** Windows 10 Corporate LTSC 64-bit, Google Chrome v98.0.4758.82
**Reproducible:** always
**Build found:** Frontend: gcp-0.0.53, Backend: gcp-0.0.112

**Preconditions**
1. Follow link https://oos.dmytrominochkin.cloud/#/
2. Log into system as a SP (e.g. 'email'="[tester2@gmail.com](mailto:tester2@gmail.com)", 'Password'="1qaz@WSX").
3. Select 'Інформація про заклад' in the menu bar at the top right corner.
4. Click on the 'Редагувати' button of the 'ІНФОРМАЦІЯ ПРО ЗАКЛАД' tab.

**Steps to reproduce**
1. Leave the 'Повна назва організації' field empty.
2. Look at the 'Далі' button at the bottom of the page.

**Actual result**
The button 'Далі' is enabled when a mandatory field is empty.
https://user-images.githubusercontent.com/97879289/153651680-3a201007-9c5d-4629-a043-d6ed54ffc381.mp4
Also is reproducible in the 'Контакти' section of the 'Редагування закладу' page.

**Expected result**
The button 'Далі' have to be disabled when a mandatory field is empty.
![AD](https://user-images.githubusercontent.com/97879289/153651693-69fa6151-f411-47ce-9c56-4da1bcafedb0.jpg)

**User story and test case links**
User story https://github.com/ita-social-projects/OoS-Backend/issues/213
Test case: [OS-293](https://jira.softserve.academy/browse/OS-293)

**Labels to be added**
Bug, pri: medium, severity: minor, UI

AnnaDidiaieva:
Verified

AnnaDidiaieva:
@Olya-web7 
**Environment:** Windows 10 Corporate LTSC 64-bit, Google Chrome v99.0.4844.74
**Reproducible**: always
**Build found:** Frontend: gcp-0.0.56, Backend: gcp-0.0.120

**Preconditions**
1. Follow link https://oos.dmytrominochkin.cloud/#/
2. Log into system as a SP (e.g. 'email'="tester2@gmail.com"), 'Password'="1qaz@WSX").
3. Select 'Інформація про заклад' in the menu bar at the top right corner.
[...]

PR #1048, opened 2022-04-10
['Редагування закладу' page] 'Далі' button stays enabled when mandatory field is empty #905 Contacts

Olya-web7:
['Редагування закладу' page] 'Далі' button stays enabled when mandatory field is empty #905


src/app/shell/personal-cabinet/provider/create-provider/create-provider.component.ts
```diff
@@ -151,8 +151,15 @@ export class CreateProviderComponent extends CreateFormComponent implements OnIn
...
-        res.push(el)
+      if ((!form.get(el).value 
+        && form.get(el).status !== 'VALID')
+        || (form?.get(el).value.street === ''
+        || form?.get(el).value.buildingNumber === ''
+        || form?.get(el).value.city === ''
+        || form?.get(el).value.district === ''
+        || form?.get(el).value.region === '')
+      ) {
+        res.push(el);
```
litvinets, line 162: you can simplify it, use for example find
and go via controls to fins the first unvalid

litvinets (changes requested):

src/app/shell/personal-cabinet/provider/create-provider/create-provider.component.ts
```diff
@@ -152,7 +152,12 @@ export class CreateProviderComponent extends CreateFormComponent implements OnIn
     let res = [];     
     for (let el in form?.controls) {
       if (!form.get(el).value && form.get(el).status !== 'VALID') {
-        res.push(el)
+        res.push(el);
+      }
+      for (let elem in form?.get(el).value) {
+        if (form?.get(el).value[elem] === '') {
+          res.push(el);
+        }
```
litvinets, line 160: I still think you can simplify it even more
So, the button should be disabled if one of the required fields is not filled
then do via the form required controls and look for the FIRST invalid/empty control. You can combine the conditions in one row. Plus, you do not need to find all invalid cases, right? If there is at least one, then it is enough for disabling. Use find/some methods

Olya-web7: fixed, pls check)

dmitryy90 (approved):

litvinets (approved):

Issue #1067, opened 2022-04-12
[Workshop Details] word declination for reviews


litvinets:
Fix word declination for reviews
![image](https://user-images.githubusercontent.com/27493959/162961810-4d170d76-d2d3-4c1e-99be-0e095e689d2c.png)

PR #1084, opened 2022-04-19
[Workshop Details] word declination for reviews

perepichai:
Task https://github.com/ita-social-projects/OoS-Frontend/issues/1067


dmitryy90 (approved):

src/app/shell/result/result.component.ts
```diff
@@ -48,6 +49,8 @@ export class ResultComponent implements OnInit, OnDestroy {
 
   isMobileScreen: boolean;
   isHidden: boolean;
+  
+  readonly Workshop = Workshop;
```
litvinets, line 53: I would rather change the naming in the component so it will be better to understand. Something like wokrshopDeclination

litvinets (approved):

Issue #928, opened 2022-02-15
[Search page] The typo in ‘Очистити фільтри’ button on the ‘Search’ page


hanna-lenko:
**Environment:** Win10 v21H2 64-bit, Chrome v 98.0.4758.82
**Reproducible:** always
**Build found:** frontend: gcp-0.0.53, backend: gcp-0.0.112

**Preconditions**
- Go to OoS (https://oos.dmytrominochkin.cloud/#/)
- Log into system as Parent (e.g. 'Email'="dub@gmail.com", 'Password'="Qwert1!")
- Go to the main page and click ‘Найпопулярніші гуртки’ hyperlink for UI verification of the ‘Search’ page.

**Steps to reproduce**
1. Verify spelling.

**Actual result**
The ‘Очистити фільтри’ button does not meet the design

![image](https://user-images.githubusercontent.com/97878413/154036170-a3618a8d-0b91-4faf-a84a-e77e7bb213df.png)

**Expected result**
The word ‘фільтри’ - all lowercase letters 
The design: https://www.figma.com/file/YRYP4zAAchOyYMgdoeP782/MoE-%7C-Dev-Team?node-id=1%3A3537 

**User story and test case links**
User story [#23](https://github.com/ita-social-projects/OoS-Backend/issues/23)
Test case [OS-260](https://jira.softserve.academy/browse/OS-260)

**Labels to be added**
"Bug", "pri: very low", "severity: trivial", Type "UI"

hanna-lenko:
The same defect in the 'Обрати вік' field

![image](https://user-images.githubusercontent.com/97878413/154064963-89405264-fcad-4ca6-9c74-186c99ae28e8.png)

litvinets:
fixede!

hanna-lenko:
Verified and closed

PR #1108, opened 2022-05-04
[General but fix] - application statuses scss fix, filter naming fix, scroll width fix

litvinets:
1) created shared scss for application statuses
2) naming fix https://github.com/ita-social-projects/OoS-Frontend/issues/928


src/app/shared/components/filters-list/filters-list.component.html
```diff
@@ -2,9 +2,9 @@
   <div fxLayout="row" fxLayoutAlign="space-between">
     <button mat-raised-button class="resetFilter" (click)="onFilterReset()">
       <mat-icon class="matFilter">delete_forever</mat-icon>
-      Очистити фільтри
+      Очистити Фільтри
```
dmitryy90, line 5: if Capiyalizing is a general rule for buttons labels you can add global style for buttons to avoid such bugs in future:
`text-transform: capitalize;`

dmitryy90 (approved):

PR #1167, opened 2022-05-13
[Application Create] - do not allowed parent to submit an application if there is an active application for the child

perepichai:
Task https://github.com/ita-social-projects/OoS-Frontend/issues/1113
Just not sure about naming.

PR #1172, opened 2022-05-16
Warning message

perepichai:
warning message and naming
https://github.com/ita-social-projects/OoS-Frontend/pull/1167#discussion_r872499814

### DIFF
--- a/src/app/shell/personal-cabinet/parent/create-application/create-application.component.ts
+++ b/src/app/shell/personal-cabinet/parent/create-application/create-application.component.ts
@@ -33,7 +33,7 @@ export class CreateApplicationComponent implements OnInit, OnDestroy {
 
   @Select(UserState.children) children$: Observable<ChildCards>;
   @Select(UserState.isAllowedNewApplication) isAllowedNewApplication$: Observable<boolean>;
-  isAllowedNewApplication: boolean;
+  isAllowedNewApplication: boolean = true;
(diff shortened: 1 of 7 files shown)


src/app/shell/personal-cabinet/parent/create-application/create-application.component.ts
```diff
@@ -53,6 +53,7 @@ export class CreateApplicationComponent implements OnInit, OnDestroy {
   destroy$: Subject<boolean> = new Subject<boolean>();
 
   ChildFormControl = new FormControl('', Validators.required);
+  isAllowed: boolean = true;
```
litvinets, line 56: You have two values for the same thing
Use isAllowedNewApplicatiom

litvinets (changes requested):

src/app/shell/personal-cabinet/parent/create-application/create-application.component.ts
perepichai: it seems I cannot use 'isAllowedNewApplication' for mat-select and small tag because the state is false by default. It means a red message will display by default. Maybe 'isAllowed' will be ok for the all cases instead of 'isAllowedNewApplicatiom'

litvinets: Then make state true by default:)

litvinets (approved):

PR #1177, opened 2022-05-19
Base img url

litvinets:


src/environments/environment.ts
```diff
@@ -6,6 +6,8 @@ export const environment = {
   production: false,
   stsServer: 'https://auth.oos.dmytrominochkin.cloud',
   serverUrl: 'https://api.oos.dmytrominochkin.cloud',
+  storageUrl: 'https://storage.googleapis.com/oos-images-63325/',
```
DmyMi, line 9: need to update prod env too :)

DmyMi (approved):

PR #1192, opened 2022-05-23
Create data module in admin-tools

Olya-web7:


litvinets (changes requested):
add routing module please as well, and add to load this module as lazy

litvinets (approved):

Issue #1146, opened 2022-05-11
[Workshop page] "Про Заклад" tab opens by default instead of "Про Гурток"


OksanaGuz:
**Environment:** macOS Big Sur 11.5.1; Chrome version 101.0.4951.54.
**Reproducible:** 

Bug reproducible only if you go second time on the workshop page.
When you open OoS first time (as register or unregister user) and click 'Детальніше' link - "Про гурток" tab opens by default.
Then if you go back to previous page and click 'Детальніше' link again - "Про заклад" tab opens. On any workshop every next time.

**Build found:**
gcp-0.0.73

**Preconditions**

1. Go to OoS as register or unregistered user (https://oos.dmytrominochkin.cloud/#)
2. Geographic location is selected (e.g. Kyiv)
3. There is at least one registered workshop (@Гурток) in the selected location

**Steps to reproduce**

1. Click the 'Детальніше' link of workshop card from any page (Найпопулярніші гуртки, Найпопулярніші напрямки, Search page list/map view, Personal cabinet)

**Actual result**

"Про заклад" tab opens by default

https://user-images.githubusercontent.com/97882565/167830137-75442c3d-781b-4880-8259-ff032897c4cf.mov

**Expected result**

"Про гурток" tab should open by default.

**User story and test case links**
User story - https://github.com/ita-social-projects/OoS-Backend/issues/98
Test case - [OS-379](https://jira.softserve.academy/browse/OS-379)

**Labels to be added**
"Bug", Priority: medium, Severity: trivial, Type UI

OksanaGuz:
@litvinets Bug is still reproducible.

litvinets:
fixed!

OksanaGuz:
Verified.

PR #1206, opened 2022-05-26
[Workshop page] "Про Заклад" tab opens by default instead of "Про Гурток"

perepichai:
Task https://github.com/ita-social-projects/OoS-Frontend/issues/1146


litvinets (changes requested):
let's think about resetting the state

PR #1239, opened 2022-05-30
[Personal cabinet Parent] The title is "Кабінет Надавача" instead of "Кабінет Користувача"

Olya-web7:
[Personal cabinet Parent] The title is "Кабінет Надавача" instead of "Кабінет Користувача"


src/app/shell/personal-cabinet/personal-cabinet.component.ts
```diff
@@ -46,6 +47,7 @@ export class PersonalCabinetComponent implements OnInit, OnDestroy {
         })
       )
     );
+    this.subrole$.subscribe((subrole: string) => this.subrole = subrole)
```
litvinets, line 50: maybe you can remove subscribe here and use async in the template? or add takeUntil to the subscription

litvinets (changes requested):

src/app/shell/personal-cabinet/personal-cabinet.component.ts
```diff
@@ -46,9 +50,11 @@ export class PersonalCabinetComponent implements OnInit, OnDestroy {
         })
       )
     );
+    this.subrole$.pipe(takeUntil(this.destroy$)).subscribe((subrole: string) => this.subrole = subrole)
```
litvinets, line 53: use async for this case please:)

litvinets (changes requested):

litvinets (approved):

Issue #1269, opened 2022-06-06
[Workshop-details] - restyle working hours view


litvinets:
URL: https://oos.dmytrominochkin.cloud/#/details/workshop/08d9c6cd-1d41-42f2-89ee-531de896f217
Change the view of working hours :
![image](https://user-images.githubusercontent.com/27493959/172147014-dba67f71-e2a4-4582-af22-762ef77c0e15.png)
New design:
Remove input borders (it is not an input, we just display hours)
Restyle selected days
![image](https://user-images.githubusercontent.com/27493959/172146914-85e0621d-9d82-41a8-b53b-bf88b669349d.png)

PR #1313, opened 2022-06-14
[Workshop-details] - restyle working hours view #1269

Pabloostr:
[Workshop-details] - restyle working hours view #1269


litvinets:
This is a shared style, it is also used for the working hours component.  Now it is a broken view for create-workshop. Revert your changes please, create separate styles for workshop-details that has no influence on other components. The view for create-workshop should remain as it was 
image
image

litvinets (changes requested):
broken view for create-workshop

litvinets (approved):
good job!

PR #1359, opened 2022-06-21
[admin-tools/provider-list] display provider info

Olya-web7:


src/app/shared/components/provider-info/provider-info.component.html
```diff
@@ -74,9 +74,9 @@ <h5>Статус</h5>
         <h5>Тип закладу</h5>
         <p class="description">{{ institutionTypes[provider.institutionType] }}</p>
         <h5>Підпорядкування</h5>
-        <p class="description">{{ provider.institution.title }}</p>
+        <p class="description">{{ provider.institution?.title }}</p>
         <h5>Про заклад</h5>
-        <p class="description">{{ provider?.description }}</p>
+        <p class="description">{{ provider.description }}</p>
```
litvinets, line 79: let's wrap it into ng-container, so we won't display title if there is not value

litvinets (changes requested):

litvinets (approved):

PR #1403, opened 2022-07-01
[Cropper] min size notification

perepichai:
What about the approach below?
image


litvinets:
@perepichai Love it!! excellent again:)!!

litvinets (approved):

litvinets (changes requested):

litvinets (approved):
fix the comment above please, I believe that condition can be simplified by ternary symbol

PR #1432, opened 2022-07-06
Create achievement fixes

Olya-web7:


src/app/shell/personal-cabinet/provider/create-achievement/create-achievement.component.html
```diff
@@ -61,32 +69,41 @@ <h3 class="wrapper-title">ДОДАТИ ДОСЯГНЕННЯ</h3>
...
         </mat-form-field>
-        <!-- <app-validation-hint
-          [validationFormControl]="AchievementFormGroup.get('description')"
-          [minCharachters]="validationConstants.INPUT_LENGTH_3"
-          [maxCharachters]="validationConstants.MAX_DESCRIPTION_LENGTH_2000">
-        </app-validation-hint> TODO: validation-->
+        <app-validation-hint 
+          [validationFormControl]="AchievementFormGroup.get('title')"
+          [isTouched]="AchievementFormGroup.get('title').touched">
+        </app-validation-hint>
```
litvinets, line 99: add the correct min/max validation (see in the section items)

litvinets:
1) add "?" 
image
2) there is a case when you select something to the multiselect and then remove it, and leave the field empty - there is no error message. try to handle this case as well
3) on create-achievement page leaving, please reset the selected workshop (there is an action for this). otherwise, when we load create-achievement for another workshops, there is a moment when the user can see the previously selected workshop
4) please, add ngIf =" workshops" for create-achievment component so we will load the page when the selected workshops is loaded as well. otherwise the page is "jumping" 
image
5) add please some margin between workshop-card and fields 
image

litvinets (changes requested):
fix comment please

litvinets (approved):

PR #1445, opened 2022-07-10
[CATEGORIES] - removed functionality regarding old version of category selection

litvinets:
1. fixed rare case issue (when the next level exists but there is no option for selection)
2. removed old categories
3. cover image is replaced with new category id

TODO:
1. remove from back-end direction, directionId, departmentId, classId
2. add filtration according to the new directions
https://github.com/ita-social-projects/OoS-Backend/issues/745


sivxkxn (approved):
great job!

Issue #859, opened 2022-02-01
[Add child] Multiple selection is not possible from the 'Cоціальна група' drop-down list


AnnaDidiaieva:
**Environment:** Windows 10 Corporate LTSC 64-bit, Google Chrome v97.0.4692.99
**Reproducible:** always (also while editing child's information)
**Build found:** Frontend: gcp-0.0.50, Backend: gcp-0.0.101

**Preconditions**
1. Follow link https://oos.dmytrominochkin.cloud/#/
2. Log into system as parent: 'Email'="dub@gmail.com", 'Password'="Qwert1!".
3. Select "Інформація про дитину" in the menu bar at the top right corner.
4. Click the "Додати дитину" button.

**Steps to reproduce**
Try to select all items from the 'Cоціальна група' drop-down list.

**Actual result**
The user can only select 1 item from the 'Cоціальна група' drop-down list.
![соц група](https://user-images.githubusercontent.com/97879289/151998566-f5d8fab3-d76e-4037-8d1f-70aaeb7808d5.jpg)

**Expected result**
Multiple selection is possible from the 'Cоціальна група' drop-down list.
![мультипл](https://user-images.githubusercontent.com/97879289/151998530-9a9369f2-dca4-41dd-aec0-9c7780d7968d.jpg)

**User story and test case links**
User story #43
Test case: https://jira.softserve.academy/browse/OS-183

**Labels to be added**
Bug, pri: high, severity: major, functional

trdmitry:
Зараз список пустий, також нема кверей на бэк /api/v1/SocialGroup/Get
![image](https://user-images.githubusercontent.com/3643368/171615410-32991551-f70d-4230-adfc-ac24878dcf1d.png)

litvinets:
the list of social groups is fixed! 
https://github.com/ita-social-projects/OoS-Frontend/pull/1261

litvinets:
![image](https://user-images.githubusercontent.com/27493959/173200318-65331175-5955-43d1-acfe-e4d4df3c991a.png)

litvinets:
if selected "ВідсутняЄ - then all options are disabled

if selected one option - it should be disabled in the dropdown - so you cannot select it again

litvinets:
design https://www.figma.com/file/YRYP4zAAchOyYMgdoeP782/MoE-%7C-Dev-Team?node-id=13404%3A7935

OksanaGuz:
Verified.

PR #1451, opened 2022-07-11
[Add child] Multiple selection is not possible from the 'Cоціальна група' drop-down list #859

Pabloostr:
#859


litvinets (changes requested):

litvinets:
1. remove required
2. checkboxes should be enabled for removing
3. fix data sending (it should be an array of objects)
4. if 'none' is selected, then only others should not be available for selection

sivxkxn (changes requested):
Data types and semicolon should be fixed

litvinets (approved):

sivxkxn (approved):
SO GOOD JOB!

Issue #1416, opened 2022-07-04
[Provider Images] - add images to the create-provider


litvinets:
Apply ​/api​/v2​/Provider​/Create
Follow the approach for Wokrshops

PR #1454, opened 2022-07-12
 [Provider Images] - add images to the create-provider #1416 

perepichai:
Task https://github.com/ita-social-projects/OoS-Frontend/issues/1416


litvinets (changes requested):
let's save the previous version (v1), so create add the same approach as for workshop. To create with or without images

litvinets (approved):

sivxkxn (approved):

Issue #1583, opened 2022-08-25
[Support/ LawsAndRegulations/ AboutPortal] - no info displaying


litvinets:
Currently, the view of no-info is styled differently:
![image](https://user-images.githubusercontent.com/27493959/186645978-0b192c00-2f1e-4e72-9fbd-f80d919f4de0.png)

![image](https://user-images.githubusercontent.com/27493959/186645726-c543b226-cc13-4e5a-a6bd-dbb3cae644fa.png)
![image](https://user-images.githubusercontent.com/27493959/186645754-d8331ecc-ca35-477f-913e-2d7d62cffa0b.png)
![image](https://user-images.githubusercontent.com/27493959/186645830-13ab0bdc-75bb-463c-ad30-009fa04bfb61.png)

To do:
1) restyle all in the same way
2) let's consider reusing the general approach for such "no-info" cases with "no-result-component'. Let's create a such banner
![image](https://user-images.githubusercontent.com/27493959/186646293-c23da4e9-d4a8-496f-bda8-d568dc0a3997.png)
![image](https://user-images.githubusercontent.com/27493959/186646117-c6959bc5-bc20-4f30-b5f0-7ee6f130bba9.png)

PR #1591, opened 2022-08-28
[Support/ LawsAndRegulations/ AboutPortal] - no info displaying

koliapatskal:
#1583 

* Added a app-no-result-card component to display a message

*Added new value NoResultTitle

*Removed css property that caused info to overlap breadcrumbs


koliapatskal:
Display information before deleting css property.

image

src/app/shell/info/about/about.component.ts
```diff
@@ -26,6 +27,7 @@ export class AboutComponent implements OnInit, OnDestroy {
 
   platformInformation: CompanyInformation;
 
+  readonly emptyList = NoResultsTitle.noInfo;
```
litvinets, line 30: this is not the proper naming, change it to more suitable such as noData/noInfo

litvinets, line 30: move read-only field to the very top (before platformInformation) so we will stay consistent we our code style. it's not a strict rule, just approach for our project

litvinets (changes requested):
spmething is wrong with you branch, therfe should be an image before the content 
image
image

litvinets (approved):

Issue #1592, opened 2022-08-28
Remove console.log from the code


litvinets:
![image](https://user-images.githubusercontent.com/27493959/187072715-01f72cbd-e8c2-4a01-b16c-5793c09a0510.png)

PR #1596, opened 2022-08-29
Remove console.log from code

koliapatskal:
#1592 

* removed all console.log commands from code


src/app/shared/services/signalR/signal-r.service.ts
```diff
@@ -29,7 +29,6 @@ export class SignalRService {
 
     this.hubConnection
       .start()
-      .then(() => console.log('Connection started'))
```
litvinets, line 32: let's save this one, so we can track successful connection, and the rest is good!

litvinets (approved):

Issue #1641, opened 2022-09-09
[Paginator state] - refactoring


litvinets:
![image](https://user-images.githubusercontent.com/27493959/189280058-ebb9c9aa-3b31-4137-8d35-2d4bce684242.png)
![image](https://user-images.githubusercontent.com/27493959/189280403-7bb8a3bd-350d-4e3f-b091-ae5b81062831.png)

let's create for the table pagination only one state.
Reason - tables always look the same on the page, so we can keep the same pagination config for it.
This does not work for directions, workshop cards, and children's cards - because the view of this functionality and the place of displaying (personal cabinet of result page) are different. 
So, please create only one state for all tables with more general naming and apply it for all tables. 

![image](https://user-images.githubusercontent.com/27493959/190628914-83bfa94c-39e0-4318-aba2-d10f443829a9.png)

friranger1:
Done!

PR #1666, opened 2022-09-21
Updated pagination #1641

friranger1:
Updated table pagination


litvinets:
1. link the issue in the header comment
2. change naming to more understandable

litvinets (commented):
doesn't work

Issue #1665, opened 2022-09-21
Apply short date pipe for data column in change log


litvinets:
![image](https://user-images.githubusercontent.com/27493959/191490127-628bc799-e9f3-4dc3-bb5c-e155c958b180.png)
apply data pipe for changes log table

| date: constants.SHORT_DATE_FORMAT

PR #1686, opened 2022-09-27
Changed short date pipe for data column #1665

friranger1:
Changed short date pipe for data column #1665


litvinets (changes requested):
you should add to ts file this constant, we should receive dd.mm.yyyy format in the end

litvinets (approved):

Issue #1715, opened 2022-10-06
[Refactoring] - move all message-bar messages to the enum/constants


litvinets:
move all messages that are needed for the snack bar to the constants/enum  (similar as what is implemented for the modal windows)
![image](https://user-images.githubusercontent.com/27493959/194295197-45dd7417-5531-4a7f-a722-25a9dfd998eb.png)
![image](https://user-images.githubusercontent.com/27493959/194295239-16d49fe3-06cc-412b-868f-98e80b3f3646.png)

PR #1720, opened 2022-10-07
move all message-bar messages to the enum

koliapatskal:
#1715 

* Created enum 'messageText'

* Replaced literals with texts from enum


src/app/shared/enum/messageBar.ts
```diff
@@ -10,3 +10,56 @@ export enum messageStatus {
   approved = 'Статус заявки успішно змінено',
   rejected = 'Заявку успішно відхилено'
 }
+
+export enum messageText{
```
litvinets, line 14: SnackbarText

Issue #1733, opened 2022-10-13
[City Confirmation] - general bugs


litvinets:
1. fix modal style
![image](https://user-images.githubusercontent.com/27493959/195613749-efb9d4f9-c86a-48b3-8fe9-2bae3b4eee1d.png)
2. three dots
![image](https://user-images.githubusercontent.com/27493959/195613979-ace54c3d-6023-410f-a60c-2ee3a7279cd8.png)
![image](https://user-images.githubusercontent.com/27493959/195614041-942a81ea-a772-4f8d-8166-8812d62ba2f8.png)
![image](https://user-images.githubusercontent.com/27493959/195614215-94558f4e-5697-45a8-a6da-63b870965200.png)
should be:
![image](https://user-images.githubusercontent.com/27493959/195614511-7fe8b6a7-2705-4bb6-9f6a-2a234f892a69.png)

3. if the city is not confirmed, then it should not be set to the local storage
![image](https://user-images.githubusercontent.com/27493959/195615147-5f569b49-3b19-4664-a81e-6bf8e5ba1380.png)

4. when you click on the "cross" then it should close the modal-confirmation-window and open the dropdown of cities from the selections 
![image](https://user-images.githubusercontent.com/27493959/195615415-0691b8f5-928f-407e-94da-fb7d5863c305.png)
![image](https://user-images.githubusercontent.com/27493959/195615530-cf12de6b-d9af-439e-95a5-c827916cd4d1.png)
![image](https://user-images.githubusercontent.com/27493959/195615856-7034482d-e359-46eb-bc08-21953c9b1a13.png)
should appear focus on this input
![image](https://user-images.githubusercontent.com/27493959/195616759-bbf935e8-f9ca-4d34-9ea4-86483a34bd08.png)

PR #1734, opened 2022-10-15
Issue#1733 city confirmation

koliapatskal:
#1733 
* Fixed cropped cross icon
* Removed wrapping of additional text about the city
* Fixed incorrectly working geolocation confirmation

### DIFF
--- a/src/app/shared/components/filters-list/city-filter/city-filter.component.html
+++ b/src/app/shared/components/filters-list/city-filter/city-filter.component.html
@@ -41,7 +41,7 @@
     </div>
     <div fxLayout="row" fxLayoutAlign="space-between center" class="city-confirmation_button">
       <button mat-button class="button" (click)="confirmCity()">Підтвердити</button>
-      <button mat-button class="button button-change" (click)="changeCity(searchInput)">Змінити</button>
+      <button mat-button class="button button-change" (click)="changeCity()">Змінити</button>
     </div>
   </div>
 </div>
--- a/src/app/shared/components/filters-list/city-filter/city-filter.component.html
+++ b/src/app/shared/components/filters-list/city-filter/city-filter.component.html
@@ -33,11 +33,11 @@
 <div class="confirmation">
   <div class="city-confirmation" *ngIf="isDispalyed && !(isConfirmCity$ | async)">
     <div fxLayout="row" fxLayoutAlign="space-between start">
-      <p>
+      <p fxFlex="93%">
         Ви знаходитесь в населеному пункті <span>{{ settlement.settlement }}</span
         >?
       </p>
-      <mat-icon (click)="isDispalyed = false">clear</mat-icon>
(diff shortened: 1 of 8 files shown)


src/app/shared/store/filter.state.ts
```diff
@@ -147,10 +147,12 @@ export class FilterState {
   constructor(private appWorkshopsService: AppWorkshopsService) {}
 
   @Action(SetCity)
-  setCity({ patchState, dispatch, getState }: StateContext<FilterStateModel>, { payload }: SetCity): void {
+  setCity({ patchState, dispatch, getState }: StateContext<FilterStateModel>, { payload, isNeedWriteInStorage }: SetCity): void {
```
litvinets, line 150: isNeedWriteInStorage  - insufficient naming, we do not need to save the value to the storage, we are doing it to confirm the city. So this value should be renamed to isConfirmedCity - if yes, then we save it to the storage

src/app/shell/shell.component.ts
```diff
@@ -22,12 +23,17 @@ export class ShellComponent implements OnInit, OnDestroy {
...
   ngOnInit(): void {
-    this.geolocationService.handleUserLocation((coords: Coords) => {
-      coords &&
-        this.geolocationService.getNearestByCoordinates(coords, (result: Codeficator) => {
-          this.geolocationService.confirmCity(result);
-        });
-    });
+    if (this.geolocationService.isCityInStorage()) {
+      this.geolocationService.confirmCity(JSON.parse(localStorage.getItem('cityConfirmation')), true);
+    } else {
```
litvinets, line 28: on init, if we don't location is forbidden and the city is not confirmed, then we should set Kyiv but not confirm it (the confirmation modal should be shown) 
image

src/app/shared/components/filters-list/city-filter/city-filter.component.html
```diff
@@ -23,24 +24,24 @@
...
         Ви знаходитесь в населеному пункті <span>{{ settlement.settlement }}</span
         >?
       </p>
-      <mat-icon (click)="isDispalyed = false">clear</mat-icon>
+      <mat-icon (click)="isDispalyed = false" fxFlex="7%">clear</mat-icon>
     </div>
     <div fxLayout="row" fxLayoutAlign="space-between center" class="city-confirmation_button">
       <button mat-button class="button" (click)="confirmCity()">Підтвердити</button>
-      <button mat-button class="button button-change" (click)="changeCity()">Змінити</button>
+      <button mat-button class="button button-change" (click)="changeCity(searchInput)">Змінити</button>
```
litvinets, line 44: I think you can get access to the search input via ViewCHild and get rid of passing the value

litvinets (changes requested):

litvinets (approved):
AMAZING JOB!!! WELL DONE, BOOOOOOOOY

Issue #1736, opened 2022-10-18
[Create-Workshop-Address] - general issues 


litvinets:
1. if the decoded address was correct, then the coordinates (long,lan) should be set to the workshop-address.
2. if the entered address does not exist on the map, then it should remove the previous marker and display "no such address is found"
![image](https://user-images.githubusercontent.com/27493959/196387524-7be37014-17e0-4db9-ad5c-483309dfa934.png)
3. if click to empty-address map place (for example river), then codeficatorid should be remained. 
![image](https://user-images.githubusercontent.com/27493959/196387997-02dbc1d7-ba0f-4c0f-a64e-89873c00d6d1.png)
4. if the address is found then should be no message about the error
![image](https://user-images.githubusercontent.com/27493959/196389624-b0aab5a5-c8be-405e-a2b4-64877eefbfe3.png)

PR #1744, opened 2022-10-19
Issue#1736 general issues map

koliapatskal:
#1736 

* Removed a command that tried to access non-existent coordinates and crashed the function
* Removed interfaces that duplicate coordinate fields from the base interface
* Fixed removal of codeficatorid when clicking on an empty place on the map
* Fixed duplicate requests when user clicks on map
* Fixed display of a search error when an address is found
* Fixed a bug where coordinates were lost due to different field names

### DIFF
--- a/src/app/shared/services/geolocation/geocoder.service.ts
+++ b/src/app/shared/services/geolocation/geocoder.service.ts
@@ -6,7 +6,6 @@ import { filter, map } from 'rxjs/operators';
 import { Geocoder } from '../../models/geolocation';
 import { GeolocationAddress } from '../../models/geolocationAddress.model';
 import { Coords } from '../../models/coords.model';
-import { GeocoderDTO } from './../../models/geolocation';
 
 @Injectable({
   providedIn: 'root',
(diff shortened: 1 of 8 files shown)


src/app/shared/services/geolocation/geocoder.service.ts
```diff
@@ -41,10 +40,10 @@ export class GeocoderService {
...
   private geocode(payload: Geocoder): Observable<Geocoder | null> {
     return this.http
-      .post<GeocoderDTO>('/api/v1/Geocoding', { ...payload, lat: payload.latitude, lon: payload.longitude })
+      .post<Geocoder>('/api/v1/Geocoding', { ...payload })
       .pipe(
-        map((result: GeocoderDTO) => {
-          return result ? { ...result, latitude: result.lat, longitude: result.lon } : null;
+        map((result: Geocoder) => {
+          return result ? result : null;
         })
```
litvinets, line 47: map((result: Geocoder) =>result ? result : null)

litvinets (approved):

Issue #1748, opened 2022-10-20
[Create Workshop] -fix institution hierarchy logic selection


litvinets:
1. Button "next" is enabled even when the institutional hierarchy is not finished yet
2. the last possible level of institution hierarchy should be set and saved to the final workshop from control
component: InstitutionHierarchyComponent, CreateDescriptionFormComponent

**_note: fix only the workshop create flow_**

PR #1753, opened 2022-10-21
Fixed institution hierarchy logic selection

koliapatskal:
#1748 
* Now only the last level of the institution's hierarchy is set and saved in the final workshop.

* When selecting new institutions, the hierarchy of controls is updated

* New controls are created gradually, not all at once


litvinets (changes requested):
There is an improvement that is needed. When I select one institution hierarchy (for example MON), then select the whole hierarchy, but after change the department again (Ministry of Sport), so the all selection was removed, but the button is still enabled 
image

litvinets (approved):

PR #1778, opened 2022-10-25
[Radius configuration] - custom radius configuration

khrMykhailo:
user-radius slider has been provided


litvinets (commented):
should be styled according to teh designs 
image

src/app/shared/components/filters-list/user-radius-set/user-radius-set.component.ts
```diff
@@ -0,0 +1,62 @@
...
+        this.store.dispatch(new GetFilteredWorkshops(true));
+        this.snackBar.open(
+          'Важливо! Тільки 100 найближчих гуртків будуть відображені на мапі.',
+          '',
+          {
+            duration: 5000,
+            panelClass: ['red-snackbar']
+          }
+        );
+      });
```
litvinets, line 60: I think you misunderstood the requirements, I don't think this message should appear each time the radius is changed, I think it should be a static display at the top of the page  (on init) with a "cross" icon, so it can be closed.
image

src/app/shared/components/filters-list/user-radius-set/user-radius-set.component.ts
```diff
@@ -0,0 +1,62 @@
...
+  extends DestroyableDirective
+  implements OnInit
+{
+  public defaultValue: number = 5;
+  public options = { floor: 2, ceil: 10 };
+  public currentRadius: FormControl = new FormControl({
+    value: '',
+    disabled: true
+  });
+  8;
```
litvinets, line 28: ?

src/app/shared/components/filters-list/user-radius-set/user-radius-set.component.ts
```diff
@@ -0,0 +1,62 @@
...
+@Component({
+  selector: 'app-user-radius-set',
+  templateUrl: './user-radius-set.component.html',
+  styleUrls: ['./user-radius-set.component.scss']
+})
+export class UserRadiusSetComponent
+  extends DestroyableDirective
+  implements OnInit
+{
+  public defaultValue: number = 5;
```
litvinets, line 22: remove typo for primitives if it i not undefined (it's a code smell)

src/app/shared/components/filters-list/user-radius-set/user-radius-set.component.ts
```diff
@@ -0,0 +1,62 @@
...
+  ngOnInit(): void {
+    this.setValue(this.defaultValue);
+    this.subscribeOnValueChanges();
+  }
+
+  public radiusHandler(event: ChangeContext): void {
+    event.pointerType && this.setValue(event.highValue);
+  }
+
+  private setValue(number): void {
```
litvinets, line 42: typo fo parameter

src/app/shared/components/filters-list/user-radius-set/user-radius-set.component.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.currentRadius.setValue(number);
+    this.store.dispatch(new SetRadiusSize(number));
```
litvinets, line 44: Cause I don't think this is relevant anymore (see my comment further), probably we don't need a separate method for setting the value this.store.dispatch(new SetRadiusSize(number));

src/app/shared/directives/destroyable.directive.ts
```diff
@@ -0,0 +1,12 @@
+import { Directive, OnDestroy } from '@angular/core';
+import { Subject } from 'rxjs';
+
+@Directive()
+export class DestroyableDirective implements OnDestroy {
+  protected readonly destroy$: Subject<void> = new Subject();
+
+  public ngOnDestroy(): void {
```
litvinets, line 8: I think piblic is redundant, it's public by default

litvinets: but I like the idea! Can you create a refactoring issue to implement such an approach all over the project?

src/app/shared/components/filters-list/filters-list.component.html
```diff
@@ -24,7 +29,12 @@
...
-    <p class="filter-text">Наявність ресурсу для надання освіти для осіб з ООП та осіб з інвалідністю</p>
-    <mat-checkbox color="primary" [formControl]="WithDisabilityOptionControl"> Наявні </mat-checkbox>
+    <p class="filter-text">
+      Наявність ресурсу для надання освіти для осіб з ООП та осіб з інвалідністю
+    </p>
+    <mat-checkbox color="primary" [formControl]="WithDisabilityOptionControl">
+      Наявні
+    </mat-checkbox>
   </div>
+  <app-user-radius-set *ngIf="isMapView"></app-user-radius-set>
```
litvinets, line 39: it's appears event for not map view
image

litvinets: Also, can you ask the designer if we can move the radius to the top? Currently, it is hidden, so we need to scroll to this opportunity, maybe at the top it will be more user friendly

litvinets: image
image

litvinets (changes requested):

src/app/shared/components/filters-list/user-radius-set/user-radius-set.component.ts
```diff
@@ -0,0 +1,62 @@
...
+import { debounceTime, takeUntil } from 'rxjs/operators';
+import { MatSnackBar } from '@angular/material/snack-bar';
+
+@Component({
+  selector: 'app-user-radius-set',
+  templateUrl: './user-radius-set.component.html',
+  styleUrls: ['./user-radius-set.component.scss']
+})
+export class UserRadiusSetComponent
+  extends DestroyableDirective
```
litvinets, line 19: don't that it is so relevant to create a separate directive that has only one default onDestroy method, it can be handled by the component, why did you decide to do it this way? 
cause it is a class that extends, so all other components can extend only one another class, but some components extend more important functionality, also some other components have another functionality on destroy, so people (students) can forget to call super.
let's keep the approach that already is applied on our project

litvinets (changes requested):

src/app/shared/components/filters-list/user-radius-set/user-radius-set.component.ts
khrMykhailo: As I understand requirements, in the input we should only watch a current number of kilometers, but not change it.

src/app/shared/components/filters-list/user-radius-set/user-radius-set.component.ts
khrMykhailo: Shure, some classes can't extend it, but the most of components haven't any class for extend, and this directive will do the code smaller.
I thought that it a good point for students a bit train with use super and overrride (with simple examples in onDestroy), so i decide to do it this way, just for some new components (not for refactoring old)
But yes, there is too many old components, so it wasn't a really good idea, i was tired, sorry)

src/app/shared/components/filters-list/user-radius-set/user-radius-set.component.ts
litvinets: I am taking about line 28

src/app/shared/components/map/map.component.ts
```diff
@@ -16,12 +23,20 @@ import { Geocoder } from './../../models/geolocation';
...
 import { FilterState } from '../../store/filter.state';
 import { SearchResponse } from '../../models/search.model';
-import { ClearCoordsByMap, SetCoordsByMap } from '../../store/filter.actions';
+import {
+  ClearCoordsByMap,
+  ClearRadiusSize,
+  SetCoordsByMap,
+  SetMapView
+} from '../../store/filter.actions';
+import { MatSnackBar } from '@angular/material/snack-bar';
```
litvinets, line 32: remove redundant import

litvinets (changes requested):

litvinets (changes requested):
image

litvinets (approved):

Issue #1768, opened 2022-10-24
[Refactroing] - rename ApplicationStatus enum to Status, apply it all over the code


litvinets:
ApplicationStatus started to be used in several places for displaying not only application status, so it should be renamed to more general naming, so it will not confuse in the code
also, probably it should be applied all over the code
(get rid of duplication, apply only one status enum)
![image](https://user-images.githubusercontent.com/27493959/197527009-11f6cb1a-d452-43eb-ba76-e37b969c27df.png)
![image](https://user-images.githubusercontent.com/27493959/197527316-d5a6de99-e49e-4cc9-97da-3f0b5a9c4eee.png)
![image](https://user-images.githubusercontent.com/27493959/197527527-8f65dd9f-fe61-437e-97bb-6fc82eef2a00.png)
![image](https://user-images.githubusercontent.com/27493959/197527582-c873feda-8b1d-467e-bae7-ee6044a47b36.png)

PR #1782, opened 2022-10-27
Issue#1768 rename application status

koliapatskal:
#1768 
* Renamed enum to more general naming
* Applied enum all over the code


src/app/shell/admin-tools/data/provider-list/provider-list.component.ts
```diff
@@ -37,12 +69,12 @@ export class ProviderListComponent implements OnInit, AfterViewInit, OnDestroy {
   readonly ownershipTypeUkr = OwnershipTypeUkr;
   readonly providerTitleUkr = ApplicationTitles;
   readonly providerAdminIcons = ApplicationIcons;
-  readonly applicationStatus = ApplicationStatus;
+  readonly applicationStatus = Statuses;
```
litvinets, line 72: it's not application status, it's provider status, so rename it please

src/app/shared/models/provider.model.ts
```diff
@@ -3,7 +3,7 @@ import { Address } from './address.model';
 import { User } from './user.model';
 import { Workshop } from './workshop.model';
 import { SectionItem } from './sectionItem.model';
-import { ApplicationStatus } from '../enum/applications';
+import { Statuses } from '../enum/applications';
```
litvinets, line 6: this is redundant import

src/app/shared/enum/applications.ts
```diff
@@ -1,4 +1,4 @@
-export enum ApplicationStatus {
+export enum Statuses {
   Pending = 'Pending',
   AcceptedForSelection = 'AcceptedForSelection',
   Approved = 'Approved',
```
litvinets, line 4: cause it's not application status anymore so it should be moved to another file (probably the new one)

litvinets: also this enum also has to be refactored  
image

litvinets (commented):
still, there are places that are missed 
image

litvinets (commented):
image
duplication should be removed, a new enum applied

litvinets (changes requested):
teh general idea - is to get rid of different "statuses" enumand keep only one, that can be applied all over the project (applications, providers, admins, etc)

src/app/shared/enum/enumUA/titles.ts
```diff
@@ -0,0 +1,28 @@
+export enum Titles {
```
litvinets, line 1: rename it to more meaningful name (for example StatusTitles)

src/app/shared/enum/enumUA/titles.ts
```diff
@@ -0,0 +1,28 @@
...
+  AcceptedForSelection = 'Прийнято до конкурсного відбору',
+  StudyingForYears = 'Навчається 1-11 років',
+  Completed = 'Навчання завершено',
+  NeverLogged = 'Очікує логування',
+  Accepted = 'Підтвердженно',
+  Blocked = 'Заблоковано',
+  Editing = 'Відправлено на редагування'
+}
+
+export enum TitlesReverse {
```
litvinets, line 16: rename to StatusTitlesReverse

src/app/shared/enum/enumUA/titles.ts
litvinets: move ti to status.ts

litvinets (approved):

Issue #1740, opened 2022-10-19
[Localization] - add functionality to switch the language, ua-en


litvinets:
when you click on the language icon, then should appear the options to switch to ua or en language
1. fix the dropdown positioning 
![image](https://user-images.githubusercontent.com/27493959/196616190-6f7ea98b-86c4-4b05-b941-e1fcb82f6a44.png)
2. the whole "basic" front-end functionality should be switched to the English/Ukrainian language; Do not consider changing the language that is an image, it will be part of the next task. 
3. data that is received from the server should not change the language 

_**note: the main idea of the task is to create the functionality to change the language, the next step will be to apply the translation for the whole website. We can start to translate only a few pages for now, for example, main-page and result page.**_ 
![image](https://user-images.githubusercontent.com/27493959/196617207-180bea0b-bcc0-4317-b2ee-37a8af8c0dfc.png)
![image](https://user-images.githubusercontent.com/27493959/196617347-f3c178b9-da17-4696-9ed3-1ca4b6e7e342.png)
![image](https://user-images.githubusercontent.com/27493959/197173271-ac5c214d-5ed3-4749-8a49-fdfa0103f8f0.png)

PR #1791, opened 2022-10-27
#1740 [Localization] 

Taras-Slovak:


src/app/header/header.component.scss
```diff
@@ -451,4 +446,14 @@
...
-}
\ No newline at end of file
+}
+
+::ng-deep {
+  .cdk-overlay-pane {
+    top: 0 !important;
+  }
+
+  // .header_lang {
```
Y-Taras, line 456: remove commented css  if not needed

src/app/header/header.component.scss
```diff
@@ -19,7 +18,8 @@
 
 .header {
   padding: 1.5rem 5rem;
-  background: url(../../assets/images/header_bg.png) center center/cover no-repeat;
+  background: url(../../assets/images/header_bg.png) center center/cover
```
Y-Taras, line 21: why do we need to put no-repeat on the next line

Issue #1788, opened 2022-10-27
[Provider Update] - update/create provider is not possible


litvinets:
investigate why provider info cannot be updated (check the creative flow as well)
![image](https://user-images.githubusercontent.com/27493959/198266530-1aa15e07-5b20-4186-87b4-0b0344483acd.png)

PR #1793, opened 2022-10-29
[Provider Update] - update/create provider is not possible

Illya97:
Follow the link: https://github.com/ita-social-projects/OoS-Frontend/issues/1788


litvinets (changes requested):
submit button should be disabled, if both checkboxes are not checked 
image
image

PR #1797, opened 2022-11-01
[Refactroing] - provide onPush strategy

khrMykhailo:
ChangeDetectionStrategy has been changed to OnPush


litvinets (approved):

litvinets (changes requested):
Let's investigate this approach a bit more

Issue #1760, opened 2022-10-24
[Chat] - Provider and Parent messaging


litvinets:
User stories:
https://github.com/ita-social-projects/OoS-Backend/issues/242
https://github.com/ita-social-projects/OoS-Backend/issues/242

To do:
1. create a chat component
![image](https://user-images.githubusercontent.com/27493959/197507371-f7938a67-836a-40ff-b48a-f423b31b17f6.png)
2. add the possibility of sending a message
3. add the possibility of displaying the history of the conversation
4. add functionality to display messages in reading/unread status

PR #1811, opened 2022-11-04
Issue#1760 messaging

koliapatskal:
#1760 
* Added chat and message components
* Added navigation for chat
* Updated chat state for receiving messages
* Implemented receiving messages through the service

Note: the ability to send a message is not developed


litvinets (approved):
add pointer on hover 
image

Issue #1690, opened 2022-09-28
[Filter List] - set selected filters to the url


litvinets:
1. on filter select, the appropriate value should be set to the url.
2. on page init filters from eth URL should be applied to the page 
Filters: 
directions,
    maxAge,
    minAge,
    isAppropriateAge,
    startTime,
    endTime,
    workingDays,
    isFree,
    isPaid,
    maxPrice,
    minPrice,
    settlement id,
    searchQuery,
    order,
    withDisabilityOption,
    isStrictWorkdays,
    isAppropriateHours,
current Page (pagination) 
![image](https://user-images.githubusercontent.com/27493959/192821537-ef78d2e0-744e-4c6b-a28c-8470af922cc0.png)

linked issue: https://github.com/ita-social-projects/OoS-Backend/issues/693

PR #1814, opened 2022-11-07
#1690 Add filter to query string

Y-Taras:


src/app/shell/result/result.component.ts
```diff
@@ -109,6 +132,22 @@ export class ResultComponent implements OnInit, OnDestroy {
     );
   }
 
+  /**
+   * Updates / Appends workshop filter to the URL query string
+   * @private
+   */
+  private setFilterStateURLParams(): void {
+    this.filterState$.pipe(takeUntil(this.destroy$)).subscribe((val) => {
+      this.filterState = val;
```
litvinets, line 141: this variable seems redundant, cannot find where it is used

src/app/shell/result/result.component.ts
```diff
@@ -109,6 +132,22 @@ export class ResultComponent implements OnInit, OnDestroy {
...
+      const filterQueryParams = Util.getFilterStateQuery(this.filterState);
+      if (filterQueryParams === '') {
```
litvinets, line 143: what do you think if we return null getFilterStateQuery if there is no such result? and then we can get rid of the code below. leave just this.router.navigate([], { queryParams: { filter: filterQueryParams }, replaceUrl: true });

litvinets (commented):
add please TODO commet for getWorkshops, because it is not used anymore or maybe even get rid of this method

src/app/shared/utils/utils.ts
```diff
@@ -218,4 +218,82 @@ export class Util {
...
+    });
+
+    return filterStateQuery;
+  }
+
+  /**
+   * Parse query string into filter state object to merge
+   * @param params - filter parameter in url query string
+   */
+  public static parseFilterStateQuery(params: string) {
```
litvinets, line 272: return typo

src/app/shell/result/result.component.ts
```diff
@@ -16,18 +17,21 @@ import { PaginatorState } from '../../shared/store/paginator.state';
...
 import { SearchResponse } from '../../shared/models/search.model';
+import { Util } from '../../shared/utils/utils';
 
 enum ViewType {
   map = 'map',
   list = 'list'
 }
 
+
+
```
litvinets, line 28: remove empty lines please

src/app/shell/result/result.component.ts
```diff
@@ -63,14 +70,30 @@ export class ResultComponent implements OnInit, OnDestroy {
...
+        }),
+        takeUntil(this.destroy$)
+      )
+      .subscribe();
+  }
 
   ngOnInit(): void {
     this.addNavPath();
-    this.getWorkshops();
+    this.store.dispatch(new GetFilteredWorkshops(this.isMapView)); // Get workshops
```
litvinets, line 89: maybe the comment should be removed?

src/app/shell/result/result.component.ts
```diff
@@ -63,14 +70,30 @@ export class ResultComponent implements OnInit, OnDestroy {
...
+      .pipe(
+        tap((params: ParamMap) => {
+          const filterParams = params.get('filter');
+          if (filterParams) {
+            this.store.dispatch(new SetFilterFromURL(Util.parseFilterStateQuery(filterParams)));
+          }
+        }),
+        takeUntil(this.destroy$)
+      )
+      .subscribe();
```
litvinets, line 84: why do you prefer to do this functionality in tap instead of subscribe? do we need this action one more time before ngOndestroy (if I get this approach correctly)?

litvinets (changes requested):

src/app/shared/utils/utils.ts
Y-Taras: typo?)

Y-Taras: https://tenor.com/view/confused-john-travolta-gif-25105508

Y-Taras: Typo is short for typographical error—a mistake made when typing something.

src/app/shell/result/result.component.ts
Y-Taras: I 'd better to leave it,  but.

src/app/shell/result/result.component.ts
Y-Taras: let us move it to subscribe. Tap is redundant

src/app/shell/result/result.component.ts
Y-Taras: I think that is good idea

Y-Taras:
> add please TODO commet for getWorkshops, because it is not used anymore or maybe even get rid of this method
missed that TODO somehow

litvinets (approved):

Issue #856, opened 2022-02-01
[Add child] Incorrect behavior is displayed when entering invalid data in the 'Дата народження' field


AnnaDidiaieva:
**Environment:** Windows 10 Corporate LTSC 64-bit, Google Chrome v97.0.4692.99
**Reproducible:** always (also while editing child's information)
**Build found:** Frontend: gcp-0.0.50, Backend: gcp-0.0.101

**Preconditions**
1. Follow link https://oos.dmytrominochkin.cloud/#/
2. Log into system as parent: 'Email'="dub@gmail.com", 'Password'="Qwert1!".
3. Select "Інформація про дитину" in the menu bar at the top right corner.
4. Click on the "Додати дитину" button.

**Steps to reproduce**
Set invalid data (1. 31/01/1902; 2. 02/02/2022; 3. ывпрвыт) in the 'Дата народження' field.

**Actual result**
1. 31/01/1902
There is no error message displayed when entering invalid data in the 'Дата народження' field.
![дата](https://user-images.githubusercontent.com/97879289/151949285-70c5e05f-3d3b-40a6-94fa-a7c5a61d4d54.jpg)
2. 02/02/2022
The uncorrect error message is displayed when entering invalid data in the 'Дата народження' field.
![Дата рождения](https://user-images.githubusercontent.com/97879289/151949304-c79c19f3-3b01-4272-9f5f-76b80e4473dd.jpg)
3. ывпрвыт
The uncorrect error message is displayed when entering invalid data in the 'Дата народження' field.
![дата рождения1](https://user-images.githubusercontent.com/97879289/151949320-d317d211-95e5-4744-bcbc-e52304fb5e61.jpg)

**Expected result**
The error message "Перевірте правильність заповнення даних" is displayed when entering invalid data in the 'Дата народження' field.
![21](https://user-images.githubusercontent.com/97879289/151951146-c3bd6587-69c8-4915-b869-b24e735281c7.jpg)
![2](https://user-images.githubusercontent.com/97879289/151951155-d8629358-54e0-405a-b781-c83bc415afbd.jpg)

**User story and test case links**
User story #43
Test case: https://jira.softserve.academy/browse/OS-186

**Labels to be added**
Bug, pri: medium, severity: minor, functional, UI

AnnaDidiaieva:
Verified

PR #1828, opened 2022-11-12
#856 Add pagination for achievements

inga-lekhman:


litvinets (changes requested):
There is no action handler in the state for SetAchievementsPerPage, so it doesn't work 
(if I set 20 items, I don't receive that 20 items and don't save this value)
image

litvinets (approved):

Issue #1835, opened 2022-11-14
[Create child] - fix page loading on refresh


litvinets:
Display a create-child form on page refreshing instead of "not page found"
![image](https://user-images.githubusercontent.com/27493959/201678883-1bba2e31-aa9a-4984-a3cb-9a1ef071664d.png)

PR #1844, opened 2022-11-15
Issue#1835 fix page loading on refresh

koliapatskal:
#1835 
* Added a new constant for modes
* Fixed 404 error for create-workshop and create-child by adding mode to parameter
* Removed unused fields and form controls
* Merged two routes redirecting to the same component into one

### DIFF
--- a/src/app/shared/constants/constants.ts
+++ b/src/app/shared/constants/constants.ts
@@ -131,6 +131,7 @@ export class CropperConfigurationConstants {
 
 export class ModeConstants {
   static readonly NEW = 'new';
+  static readonly WORKSHOP = 'workshop';
 }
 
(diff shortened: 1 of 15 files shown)


litvinets (changes requested):
check the same for provider-admin , admin ministry,

src/app/shared/constants/constants.ts
```diff
@@ -129,4 +129,8 @@ export class CropperConfigurationConstants {
   };
 }
 
+export class ModeConstants {
+  static readonly NEW = 'new';
```
litvinets, line 133: add to the mode constants companion and other possible param that are used

litvinets (approved):

Issue #1820, opened 2022-11-08
[Create workshop] - add info hint


litvinets:
credentials:
Надавач (Директор школи)
Логін: sleep@gmail.com
Пароль: A12345a@
user role: provider

url: http://localhost:4200/#/create-workshop/

![image](https://user-images.githubusercontent.com/27493959/200870497-1e5ecec6-1ad7-499d-b1dc-4665f07dd33b.png)

similar icon --> ApplicationCardComponent
Related user story: https://github.com/ita-social-projects/OoS-Backend/issues/820

PR #1851, opened 2022-11-16
#1820 create workshop info hint

nazarstig:


litvinets:
should be centered 
image

litvinets (changes requested):

litvinets (approved):

Issue #1829, opened 2022-11-14
[Refactoring] - improve http params


litvinets:
Apply the approach for setting params (at least for pagination)
![image](https://user-images.githubusercontent.com/27493959/201616918-7d979878-98cd-480b-b1d8-9630df1dcf90.png)

PR #1853, opened 2022-11-17
Implemented chain setting for HttpParams

koliapatskal:
#1829 
* Implemented chain setting for HttpParams


litvinets (approved):
well done

PR #1854, opened 2022-11-17
Issue#1839 - empty cell handling

koliapatskal:
#1839 
* Added pipe to get full address
* Added a pipe to process the value for emptiness
* Applied a pipe to all parts of the code where an empty value can be present


litvinets (changes requested):
apply this pipe to other tables

litvinets (approved):

Issue #1805, opened 2022-11-03
[Create provider] - remove divider line for add new section


litvinets:
Remove the divider for the new section, it should be added only between previous sections
(check the same for create-workshop)

components: CreatePhotoFormComponent, CreateDescriptionFormComponent
![image](https://user-images.githubusercontent.com/27493959/199688501-d1063890-2fdf-412b-908a-64b58d853fcf.png)
![image](https://user-images.githubusercontent.com/27493959/199688549-292a50f1-48eb-430d-8a10-139ebd513c8b.png)

PR #1888, opened 2022-12-05
#1805 added check on index for border style

nazarstig:


litvinets (changes requested):
apply please same approach to create-teacher logic and create-children
image
image

src/app/shared/components/info-form/info-form.component.html
```diff
@@ -3,7 +3,7 @@
   fxLayout="column"
   fxLayoutAlign="center space-between"
   class="step"
-  [ngClass]="formAmount > 1 ? 'step-border' : ''"
+  [ngClass]="(formAmount > 1) && (index !== formAmount - 1) ? 'step-border' : ''"
```
khrMykhailo, line 6: mb [class.step-border]="condition" would be better then ngClass

litvinets (approved):

PR #1896, opened 2022-12-12
1825 Create new translation pipe for cases

inga-lekhman:


litvinets (changes requested):
image
probably teh case where it should be "діти"

litvinets:
image
same here, it should be workshops/гуртки

litvinets (changes requested):
unfortunately it does not work, for case when it should be "workshops" it has "children instead (for UA as well)
image
image

litvinets (approved):

PR #1901, opened 2022-12-14
Localized create-child form

koliapatskal:
* Localized create-child form


litvinets (changes requested):
In scope of thsi PR pelase, can you provide some fixes? 
1. remove this redundant space please 
image
2. Remove button should be disabled if there is only one form in this form array, so we son't have such case: 
image
image
3. text with checkbox should be centered (check how it is implemented in create-application) 
image
4. confirmation message should have "?" sign in the end, apply everwhere 
image

litvinets (approved):
well done! I have provided small changes (the remove button is hidden is there is only one form in the array, I think it is more user-friendly. The user is not confused why this button is disabled)

The rest works perfect, thanks!

Issue #1905, opened 2022-12-14
[Admin Tools] - do not display the table until data is loaded


litvinets:
> do not display the table until data is loaded
> if data is loaded and there is none, then display no-info banner

![image](https://user-images.githubusercontent.com/27493959/207625638-60d31156-f451-49da-95a7-ba904cc2bde4.png)

PR #1908, opened 2022-12-15
Issue 1905 - wrong table data display

koliapatskal:
#1905 

* Added condition for displaying the table
* Added info card in case of missing data from the server


litvinets (approved):
1) I  have removed redundant action and API request that is not used in the application anymore. 
2) I have fixed the provider search interface that was wrong implemented in the state, and fixes no result response. Before it was displayed the table even when there were no results

Issue #1900, opened 2022-12-14
[New workshop] There remains 'Це поле є обов'язковим' error after choosing a day in the schedule of classes.


ViktoriiaHrytchyna:
**Environment:** Windows 10, 64-bit, Google Chrome 107.0.5304.107.
**Reproducible:** always.

**Preconditions**
1.Follow the link https://oos.dmytrominochkin.cloud/#/
2.Log in as SP (e.g. 'email'="gadigob939@probdd.com", 'Password'="Qazwsx789!")
3.Select «Мої гуртки» in the menu bar at the top right corner.
4.Click on the “Додати гурток” button and fill in all mandatory fields.

**Steps to reproduce**
1. Select Monday and Thursday in the schedule.
2. Cancel the choice of days.
3. Select again Monday.
4. Pay attention to the 'Це поле є обов'язковим' error.

**Actual result**
There remains 'Це поле є обов'язковим' error after choosing a day in the schedule of classes.
![Снимок экрана 2022-12-12 225051](https://user-images.githubusercontent.com/115226563/207548022-1e31be1a-281e-451b-acf0-aeb3ab10f856.jpg)

**Expected result**
After choosing a day in the schedule of classes, the error disappears.

**Labels to be added**
"Bug", Priority ("low "), Severity ("low").

ViktoriiaHrytchyna:
Bug is fixed, thank you)))

PR #1913, opened 2022-12-16
Issue#1900 - Added manual change detection

koliapatskal:
#1900 

* Fixed an issue where changes were not detected and the old value was used


litvinets (changes requested):
provider onPush strategy please for this component

litvinets (approved):

PR #1992, opened 2023-01-24
Updated admins pagination

koliapatskal:


litvinets:
@koliapatskal  let's go with another approach. thanks for raising this PR as well

PR #1994, opened 2023-01-24
[LOCALIZATION] - parent view

litvinets:
1. user config localization
2. empty banner localization
3. placeholders localization


koliapatskal (approved):
"Статуси заявок" not localized
image

PR #1996, opened 2023-01-25
[LOCALIZATION] - provider view

litvinets:
create provider form and support components were localized
create-provider-admins
providers-admins

NOTE: users-list component is not localized yet!


koliapatskal (approved):
misspelled word

Unformation -> Information
image

PR #1997, opened 2023-01-26
[LOCALIZATION] - nav bar, admin tools, side nav bar

litvinets:


koliapatskal (changes requested):
No-notification message no localized
image

notification actions no localized
image

Wrong localize path on providers tab(admin-tools)
image

Direction placeholder not localized(admin-tools)
image

'Адміністрування' title no localized
image

koliapatskal (approved):

PR #1998, opened 2023-01-30
Favorite workshops - pagination

litvinets:


koliapatskal (changes requested):
No message about missing items
image

An array of 8 elements comes from the server, although 12 is specified in the request. Please look at what is the problem
image
image

koliapatskal (approved):

Issue #1817, opened 2022-11-08
[Block/Unblock provider] - tech admin block/unblock providers


litvinets:
related issue: https://github.com/ita-social-projects/OoS-Backend/issues/868
add functionality to block/unblock providers for tech admin role
![image](https://user-images.githubusercontent.com/27493959/200554647-5e4eb891-7f5e-440e-8dce-8eb44a47b80e.png)

PR #2000, opened 2023-01-30
Reinvitation/ Block/ Unblock of provider admins, ministry admins, provider

litvinets:
1. send re-invitation for provider admin
2. block unblock ministry admin
3. block unblock provider
4. block unblock provider admin

https://github.com/ita-social-projects/OoS-Frontend/issues/1817


koliapatskal (approved):
Should a modal window appear to indicate the reason for blocking a provider-admin? If yes, then it doesn't happen.

litvinets:
@koliapatskal , no, there should be no reason for blocking (no field i=for payload, no requirements)

PR #2022, opened 2023-02-09
Issue#2018-Fixed property access for undefined variable

koliaBp:
#2018 

* Fixed a bug when a property was accessed on a undefined object

### DIFF
--- a/src/app/shared/services/children/children.service.ts
+++ b/src/app/shared/services/children/children.service.ts
@@ -21,7 +21,7 @@ export class ChildrenService {
         params = params.set('SearchString', parameters.searchString);
       }
 
-      if (parameters.isParent !== undefined) {
+      if (parameters.isParent !== null) {
         params = params.set('isParent', parameters.isParent.toString());
       }
     }
--- a/src/app/shared/services/children/children.service.ts
+++ b/src/app/shared/services/children/children.service.ts
@@ -23,7 +23,7 @@ export class ChildrenService {
         params = params.set('SearchString', parameters.searchString);
       }
 
-      if (typeof parameters.isParent !== 'boolean') {
+      if (typeof parameters.isParent === 'boolean') {
         params = params.set('isParent', parameters.isParent);
       }
     }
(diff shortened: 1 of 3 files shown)


src/app/shared/services/children/children.service.ts
```diff
@@ -21,7 +21,7 @@ export class ChildrenService {
         params = params.set('SearchString', parameters.searchString);
       }
 
-      if (parameters.isParent !== undefined) {
+      if (parameters.isParent !== null) {
```
humanmantis, line 24: It may be better to leave both conditions here because if the value is `undefined`, the same error will occur
At least in [one place](https://github.com/ita-social-projects/OoS-Frontend/blob/c739f8c35c9a3a285f2489b98edecba062533ed0/src/app/shell/admin-tools/data/users/users.component.ts#L94), we assign `undefined` to `isParent`

humanmantis (changes requested):

src/app/shared/services/children/children.service.ts
```diff
@@ -1,28 +1,30 @@
...
     let params = new HttpParams();
 
     if (parameters) {
       if (parameters.searchString) {
         params = params.set('SearchString', parameters.searchString);
       }
 
-      if (parameters.isParent !== undefined) {
-        params = params.set('isParent', parameters.isParent.toString());
+      if (typeof parameters.isParent !== 'boolean') {
```
humanmantis, line 26: The `isParent` can be `boolean` or `null` / `undefined` (possibly).
I don't think it makes sense to pass `isParent=null` to the server, you should pass only this property when the type is `boolean`, and not vice versa.

humanmantis (changes requested):

src/app/shared/services/children/children.service.ts
koliapatskal: I agree, I got confused in the condition. I'll fix it)

humanmantis (approved):

Issue #1979, opened 2023-01-20
[Localization] - localize date and calendar


litvinets:
add localization for the calendar and all dates
![image](https://user-images.githubusercontent.com/27493959/213688783-b8725fa9-1e08-423c-b60f-6b88d2bf785f.png)
![image](https://user-images.githubusercontent.com/27493959/213688817-49e8168c-42f4-4862-bc47-db3e62e4d2b6.png)
![image](https://user-images.githubusercontent.com/27493959/213688886-75083d88-b1b3-4aa3-b1d5-1980f74d6791.png)
![image](https://user-images.githubusercontent.com/27493959/213689055-89aa1cf2-7d6c-4207-aa52-68a567997231.png)

PR #2023, opened 2023-02-09
Issue#1979 localize calendar

koliaBp:
#1979 

* Added localization change for datepicker
* Added remembering the selected localization

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -1,43 +1,45 @@
-import { ParentState } from './shared/store/parent.state.';
-import { BrowserModule } from '@angular/platform-browser';
-import { NgModule } from '@angular/core';
-import { AppRoutingModule } from './app-routing.module';
-import { AppComponent } from './app.component';
-import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
-import { HeaderComponent } from './header/header.component';
-import { ShellModule } from './shell/shell.module';
+import { registerLocaleData } from '@angular/common';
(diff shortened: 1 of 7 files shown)


src/app/app.module.ts
```diff
@@ -37,9 +35,9 @@ import { ChatState } from './shared/store/chat.state';
 import { HttpClientModule, HttpClient } from '@angular/common/http';
 import { TranslateModule, TranslateLoader } from '@ngx-translate/core';
 import { TranslateHttpLoader } from '@ngx-translate/http-loader';
+import { DateAdapter } from '@angular/material/core';
```
humanmantis, line 38: It is recommended to sort the imported files in the following order:

1. angular libs
2. external libraries
3. internal files

You can use [this extension for VSCode](https://marketplace.visualstudio.com/items?itemName=mike-co.import-sorter) to fix the following errors more easily

src/app/app.module.ts
```diff
@@ -97,9 +95,18 @@ registerLocaleData(localeUk);
   bootstrap: [AppComponent]
 })
 export class AppModule {
-  constructor() {
-    localStorage.setItem('ui-culture', 'uk');
+  constructor(private dateAdapter: DateAdapter<Date>) {
+    dateAdapter.setLocale(selectedLanguage);
```
humanmantis, line 99: Can we move this to app.component? 
It is better to avoid adding logic to the module.

humanmantis (changes requested):

humanmantis (approved):

Issue #2017, opened 2023-02-06
[Localization] - history logs switch localization languages


litvinets:
for history changes table 0> language should be switched (for Ukrainian version there is English text)
![image](https://user-images.githubusercontent.com/27493959/216986013-d82fb101-789d-48d6-b25a-bcecbab4be0e.png)

PR #2029, opened 2023-03-01
History logs switch localization languages

koliapatskal:
#2017 

* Fixed incorrect attribute name, on which the choice of translation depended
* Fixed a bug where we tried to access the request response property before getting the response itself
* Moved subscriptions to observables to the ngOnInit method, removing the use of take(1) operators


src/app/shell/admin-tools/data/history-log/history-log.component.ts
```diff
@@ -1,28 +1,32 @@
+import { Observable, Subject } from 'rxjs';
+import {
+    debounceTime, distinctUntilChanged, filter, map, startWith, takeUntil
```
humanmantis, line 3: Looks like there are 4 spaces instead of 2.
If you use an extension to sort imports, add this to the settings config file
`"importSorter.importStringConfiguration.tabSize": 2`

src/app/shell/admin-tools/data/history-log/history-log.component.ts
```diff
@@ -80,6 +84,21 @@ export class HistoryLogComponent implements OnInit, OnDestroy {
...
+      .subscribe((providerHistories: SearchResponse<ProviderHistory[]>) => (this.totalAmount = providerHistories.totalAmount));
+    this.providerAdminHistory$
+      .pipe(filter(Boolean), takeUntil(this.destroy$))
+      .subscribe(
+        (providerAdminHistories: SearchResponse<ProviderAdminHistory[]>) => (this.totalAmount = providerAdminHistories.totalAmount)
+      );
+    this.applicationHistory$
+      .pipe(filter(Boolean), takeUntil(this.destroy$))
+      .subscribe((applicationHistories: SearchResponse<ApplicationHistory[]>) => {
+        console.log(applicationHistories);
```
humanmantis, line 99: remove `console.log`

humanmantis (changes requested):

humanmantis (approved):

Issue #2013, opened 2023-02-06
[Provider-admin / Deputy] - managed workshop list displaying


litvinets:
1. For provider-admin-provider-deputy display workshop-checkbox selection for filtering (application page)

ProviderApplciationsComponent, method getProviderWorkshops
for **subRole === providerAdmin** , use a new endpoint 
![image](https://user-images.githubusercontent.com/27493959/216914892-24a830b8-def9-4e4d-ab3d-e462c80804e7.png)

_note: deputy and provider (none subrole) receive the same workshop list, so I else case should be only providerAdmin_
Current:
![image](https://user-images.githubusercontent.com/27493959/216914124-969d6e19-19c2-48ba-a3ff-f8361fca171f.png)
Should be:
![image](https://user-images.githubusercontent.com/27493959/216914260-09a59be9-2f4a-4a1a-a078-ea1150adde96.png)

2. for provider-admin edit -> check already assigned workshops 
use the same endpoint to get assigned workshop in edit mode for provider-admin
![image](https://user-images.githubusercontent.com/27493959/216915323-40eae1c8-3934-481b-9ba1-8db3aa501cf7.png)
![image](https://user-images.githubusercontent.com/27493959/216916398-5e37c504-0885-4b21-84c4-003b373dc97e.png)
get assigned workshop list and check the same workshops in the dropdown (**ONLY PROVIDER_ADMIN SUBROLE**)
![image](https://user-images.githubusercontent.com/27493959/216916634-8cccbe17-2daa-4095-9fe0-7be9c3672c28.png)

back-end changes: https://github.com/ita-social-projects/OoS-Backend/pull/985

**Credentials:**

ProviderAdmin
Login: sorigo8221@cnogs.com , 
Password: sv!sc?9F

Deputy
Login: narigot790@ulforex.com
Password: 0E-wHhEJ

PR #2035, opened 2023-03-15
Issue#2013 managed workshops lists

koliapatskal:
#2013 

* Added a new endpoint for getting a list of workshops
* Now the role of the providerDeputy will receive the list of workshops and applications in the same way as the role of the provider
* Added logic for marking all workshops for providerAdmin on edit page
* Fixed a bug when sending null to OnUpdateProviderAdminSuccess, where they tried to access its properties. This happened because we expected the request to return the providerAdmin, however it does not return anything
* Updated the data model of the providerAdmin for the changed model from the server


src/app/shell/personal-cabinet/provider/create-provider-admin/create-provider-admin.component.ts
```diff
@@ -121,12 +122,22 @@ export class CreateProviderAdminComponent extends CreateFormComponent implements
...
+    combineLatest([this.providerAdmin$, this.truncatedItems$])
+      .pipe(
+        filter(([providerAdmin, allEntities]: [ProviderAdmin, TruncatedItem[]]) => !!(providerAdmin && allEntities)),
+        takeUntil(this.destroy$)
+      )
+      .subscribe(([providerAdmin, allEntities]) => {
+        this.ProviderAdminFormGroup.patchValue(providerAdmin, { emitEvent: false });
+        this.entityControl.setValue(
+          allEntities.filter((entity: TruncatedItem) => {
+            const isChecked = !!providerAdmin.workshopTitles.filter((checkedEntity) => entity.id === checkedEntity.id).length;
```
humanmantis, line 134: `filter` can be replaced by `find`.
we can remove `isChecked` variable and just return value like `() => items.find(...)`

humanmantis (changes requested):
Just in general. It's a good practice to add an access modifier (public/protected/private) to every variable and method in a class. 
I think we should add it at least in the new code (for now).

humanmantis (approved):

PR #2041, opened 2023-03-22
Fixed favorites pagination

koliapatskal:
* Fixed a bug when switching to the next page and switching the number per page did not work


humanmantis (changes requested):
Don't forget about access modifiers for fields and methods

humanmantis (approved):

Issue #1978, opened 2023-01-20
[Workshop] There is no an error message when the application limit is exceeded when the Parent applies to the workshop for his children.


ViktoriiaHrytchyna:
**Environment:** Windows 10, 64-bit, Google Chrome 107.0.5304.107.
**Reproducible:** always.

**Preconditions**
1.Go to link https://oos.dmytrominochkin.cloud/#/ as registered user, Parent role (Login: 'pidim28018@irebah.com'  Password: 'Qazwsx789!')
2.At least one child is created in the Parent`s profile.
3.The location is selected (e.g. Kyiv)
4.There is at least one workshop registered by Service provider in selected location (Login: 'lidoyo6572@cmeinbox.com'  Password: 'Qazwsx789!').
5.A parent hasn't applied to chosen workshops within week before test execution.
6.The interested workshop is found and the 'Workshop' page is opened (by clicking the "Детальніше" button of the workshop card).
7.The 'Application' page is opened (by clicking the 'Подати заяву' button).
8.The option "Подати дані про дитину" is chosen. 
9.The child is chosen from "Виберіть дитину" dropdown list AND all checkboxes are selected.

**Steps to reproduce**
1. Submit an application to the workshop for the first time by clicking 'Відправити' button on the 'Application' page and then the 'Підтвердити' button in the confirmation popup.
2. Log in as SP and change status "Очікує підтвердження" on the "Відмовлено" status.
3. Log in as Parent.
4. Submit an application to the same workshop for the same child for the second time.
5. Log in as SP and change status "Очікує підтвердження" on the "Відмовлено" status.
6. Log in as Parent.
7. Submit an application to the same workshop for the same child for the third time.
8. Pay attention to the absence of a notification that the limit has been exceeded.

**Actual result**
There is no notification when the application limit is exceeded when the Parent applies to the workshop for his children.
https://drive.google.com/file/d/1Yr-51FZd_YV1VIsk9DCrhAv38KmqZy8_/view?usp=share_link

**Expected result**
An error message appears on the screen: "Перевищено ліміт заявок. Спробуйте ще раз через тиждень". The application is not submitted.

**User story and test case links**
https://jira.softserve.academy/browse/OS-431

**Labels to be added**
"Bug", Priority ("medium"), Severity ("medium"), Type ("UI, "Functional").

ViktoriiaHrytchyna:
Fixed

PR #2046, opened 2023-03-31
Issue#1978 application limit

koliapatskal:
#1978 

* Updated interceptor operation, now after processing an error it rolls it on, instead of creating a new error
* The error handling logic has been changed. Previously, the error type was determined by the message text that the interceptor transmitted. Now the error is determined by its status.
* The message-bar styles have been updated. The text is now placed in the body of the notification

### DIFF
--- a/src/app/shared/store/admin.state.ts
+++ b/src/app/shared/store/admin.state.ts
@@ -1,97 +1,56 @@
-import { Action, Selector, State, StateContext, Store } from '@ngxs/store';
-import { Direction } from '../models/category.model';
-import { MarkFormDirty, ShowMessageBar } from './app.actions';
 import { Observable, of } from 'rxjs';
 import { catchError, tap } from 'rxjs/operators';
-import { DirectionsService } from '../services/directions/directions.service';
-import { Child } from '../models/child.model';
-import { ChildrenService } from '../services/children/children.service';
-import { CompanyInformation } from '../models/сompanyInformation.model';
-import { Injectable } from '@angular/core';
-import { Parent } from '../models/parent.model';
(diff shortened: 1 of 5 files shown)


src/app/shared/store/admin.state.ts
```diff
@@ -540,7 +496,7 @@ export class AdminState {
 
   @Action(OnCreateMinistryAdminFail)
   onCreateMinistryAdminFail({ dispatch }: StateContext<AdminState>, { payload }: OnCreateMinistryAdminFail): void {
-    const message = payload.message === 'Internal Server Error' ? SnackbarText.errorEmail : SnackbarText.error;
+    const message = payload.status === 500 ? SnackbarText.errorEmail : SnackbarText.error;
     dispatch(
       new ShowMessageBar({
         message: message,
```
humanmantis, line 502: can be just `message`, we can use shorthand property syntax

src/app/shared/store/admin.state.ts
```diff
@@ -838,7 +774,7 @@ export class AdminState {
 
   @Action(OnCreateRegionAdminFail)
   onCreateRegionAdminFail({ dispatch }: StateContext<AdminState>, { payload }: OnCreateRegionAdminFail): void {
-    const message = payload.message === 'Internal Server Error' ? SnackbarText.errorEmail : SnackbarText.error;
+    const message = payload.status === 500 ? SnackbarText.errorEmail : SnackbarText.error;
     dispatch(
       new ShowMessageBar({
         message: message,
```
humanmantis, line 780: same here

humanmantis (changes requested):

humanmantis (approved):

PR #2047, opened 2023-03-31
Issue#301 tech admin add delete rating

OlyaHlova:


src/app/shell/details/details-tabs/reviews/rate/rate.component.html
humanmantis, line 1: There is problem with EOL in last row

humanmantis (changes requested):
It would be better to add an access modifier to the methods and properties of our classes. 
This will improve readability and make the code a little more understandable in general. 
Ignore state for now.

humanmantis (approved):

Issue #2127, opened 2023-06-21
[Personal cabinet] Redirect to the wrong page after canceling editing of the "Особиста інформація" page.


ViktoriiaHrytchyna:
**Environment:** Windows 10, 64-bit, Google Chrome 107.0.5304.107.
**Reproducible:** always.

**Preconditions**
1.Go to https://oos.dmytrominochkin.cloud/#/
2.Log in as service provider (email: jigoc25090@klblogs.com,  pass: Qazwsx789!).

**Steps to reproduce**
1. Go to the "Особиста інформація" tab of personal cabinet (by clicking the button in the top right corner).
2. Click the "Редагувати" button.
3. Change some data.
4. Cancel the edit process by clicking the "Скасувати" button.
5. Click the "Скасувати" button in pop-up message.
6. Repeat step 4 (Click the "Скасувати" button).
7. Click the "Підтвердити" button in pop-up message.

**Actual result**
At the second cancellation of editing, a redirection occurs to the page that preceded the entry to the "Особиста інформація" page.

**Expected result**
After canceling editing, you will be redirected to the "Особиста інформація" page.

**Labels to be added**
"Bug", Priority ("pri: low "), Severity ("severity: low"), Type ("Functional").

ViktoriiaHrytchyna:
Fixed

PR #2132, opened 2023-06-27
Oliinyk / 2127 Fix redirect after canceling dirty form

doliinyk:
#2127 

Bug is caused by using `this.location.back()` in components with such edit/create forms and with CreateGuard.
The reason is that `location` navigates through navigation story regardless of CreateGuard approach. So, if user will change some info in such forms, try to cancel, reject canceling and then trying to cancel once more and then confirm canceling, navigation tree will be shifted as times as user rejected cancel.

### DIFF
--- a/src/app/shell/admin-tools/data/directions-wrapper/directions/create-direction/create-direction.component.ts
+++ b/src/app/shell/admin-tools/data/directions-wrapper/directions/create-direction/create-direction.component.ts
@@ -54,7 +54,7 @@ export class CreateDirectionComponent extends CreateFormComponent implements OnI
 
   ngOnInit(): void {
     this.determineEditMode();
-    this.addNavPath(); // TODO: move this to abstract create-form component
+    this.addNavPath();
   }
 
   addNavPath(): void {
--- a/src/app/shell/admin-tools/data/directions-wrapper/directions/create-direction/create-direction.component.ts
+++ b/src/app/shell/admin-tools/data/directions-wrapper/directions/create-direction/create-direction.component.ts
@@ -78,6 +78,7 @@ export class CreateDirectionComponent extends CreateFormComponent implements OnI
   }
 
   onBack(): void {
+    // TODO: Fix redirection when edit canceled onBack() and then confirm
     this.location.back();
   }
 
--- a/src/app/shell/admin-tools/data/directions-wrapper/directions/create-direction/create-direction.component.ts
+++ b/src/app/shell/admin-tools/data/directions-wrapper/directions/create-direction/create-direction.component.ts
@@ -1,22 +1,22 @@
-import { CreateDirection, GetDirectionById, UpdateDirection } from './../../../../../../shared/store/admin.actions';
-import { takeUntil, filter } from 'rxjs/operators';
 import { STEPPER_GLOBAL_OPTIONS } from '@angular/cdk/stepper';
 import { Component, OnDestroy, OnInit } from '@angular/core';
-import { ActivatedRoute } from '@angular/router';
+import { FormBuilder, FormControl, FormGroup, Validators } from '@angular/forms';
+import { MatDialog } from '@angular/material/dialog';
+import { ActivatedRoute, Router } from '@angular/router';
 import { Select, Store } from '@ngxs/store';
 import { Observable } from 'rxjs';
-import { Location } from '@angular/common';
-import { FormGroup, FormBuilder, FormControl, Validators } from '@angular/forms';
-import { MatDialog } from '@angular/material/dialog';
-import { ConfirmationModalWindowComponent } from '../../../../../../shared/components/confirmation-modal-window/confirmation-modal-window.component';
-import { Constants } from '../../../../../../shared/constants/constants';
-import { ModalConfirmationType } from '../../../../../../shared/enum/modal-confirmation';
-import { NavBarName } from '../../../../../../shared/enum/enumUA/navigation-bar';
-import { Direction } from '../../../../../../shared/models/category.model';
-import { NavigationBarService } from '../../../../../../shared/services/navigation-bar/navigation-bar.service';
-import { AdminState } from '../../../../../../shared/store/admin.state';
-import { AddNavPath } from '../../../../../../shared/store/navigation.actions';
+import { filter, takeUntil } from 'rxjs/operators';
+
+import { ConfirmationModalWindowComponent } from 'shared/components/confirmation-modal-window/confirmation-modal-window.component';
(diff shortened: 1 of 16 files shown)


src/app/shell/personal-cabinet/shared-cabinet/user-config/user-config-edit/user-config-edit.component.ts
```diff
@@ -124,6 +123,11 @@ export class UserConfigEditComponent extends CreateFormComponent implements OnIn
   }
 
   public onCancel(): void {
-    this.location.back();
+    this.navigationPaths$
+      .subscribe((value) => {
+        const path = value[value.length - 2].path ?? '/personal-cabinet/config';
+        this.router.navigate([path]);
+      })
+      .unsubscribe();
```
doliinyk, line 131: Other solution is using hard navigation to parent page like it is in CreateProviderComponent:
`this.router.navigate(['/personal-cabinet/provider/info']);`
In our case it will look like:
`this.router.navigate(['/personal-cabinet/config]);`

DmyMi: maybe we can play with this or the old school location back in [ConfirmationModalWindowComponent](https://github.com/ita-social-projects/OoS-Frontend/blob/develop/src/app/shared/components/confirmation-modal-window/confirmation-modal-window.component.ts).
Something like if type of modal is `ModalConfirmationType.leavePage` then the confirm button on the modal will navigate back. That way we probably can modify only the modal code so it will work in the entire project and think about some other option to trigger the candeactivate guard in `onBack()` instead of `location.back()`

doliinyk: Do we actually need to use `location.back()`? Maybe we should make navigation like in [CreateProviderComponent](src/app/shell/personal-cabinet/provider/create-provider/create-provider.component.ts) - just to parent route?
I think no one is going to directly navigate to form create/edit URL for example from other website, edit info and then return to that website instead of seeing results of actions.
Anyway, I haven't found approaches to manually trigger `CreateGuard.canDeactivate()` without routing, also there is option for injecting `CreateGuard`, but it's many code for each component with form. I'm suggesting use `this.router.navigate[]` for parent routes

humanmantis: As for me, I see 2 possible solutions:
1) use a direct route like `this.router.navigate(['/personal-cabinet/config]);`
2) since `onCancel` or `onBack` are used in many forms, you can try to figure out a more general solution, it could be something like this `this.router.navigate(['../'], { relativeTo: this.route });`, but it requires more local testing to make sure everything works as expected. 

So if you have time, you can test the second option for other cases, or we can replace `this.location.back()` with `this.router.navigate(['/personal-cabinet/config]);`.

humanmantis (changes requested):

src/app/shell/personal-cabinet/shared-cabinet/user-config/user-config-edit/user-config-edit.component.ts
doliinyk: > As for me, I see 2 possible solutions:
> 
> 1. use a direct route like `this.router.navigate(['/personal-cabinet/config]);`
> 2. since `onCancel` or `onBack` are used in many forms, you can try to figure out a more general solution, it could be something like this `this.router.navigate(['../'], { relativeTo: this.route });`, but it requires more local testing to make sure everything works as expected.
> 
> So if you have time, you can test the second option for other cases, or we can replace `this.location.back()` with `this.router.navigate(['/personal-cabinet/config]);`.

What about using `routerLink` attribute in HTML template for components? I have tested it locally with route `'../'` - it doesn't break navigation history and calling `CreateGuard` well. Since `onCancel` or `onBack` methods don't have any additional logics (in cases where they are) I suggest using `routerLink`, what do you think?
Also I have noted that some paths in create/update forms don't using route as parent component. For example route `/create-admin` for `CreateAdminComponent` and `'/admin-tools/info'` path that calls form, so route `'../'`. will not work correctly for that component and the option is to specify `'/admin-tools/info'`.

humanmantis: The `routerLink` does the same thing as `this.router.navigate`, so both solutions are good. 
I prefer the solution of keeping the `onCancel` method, and we can add it as an abstract to `CreateFormComponent` (and `onSubmit` too) because in every form that implements this interface we have this logic.

src/app/shell/admin-tools/data/directions-wrapper/directions/create-direction/create-direction.component.ts
```diff
@@ -1,22 +1,22 @@
...
+import { filter, takeUntil } from 'rxjs/operators';
+
+import { ConfirmationModalWindowComponent } from 'shared/components/confirmation-modal-window/confirmation-modal-window.component';
+import { Constants } from 'shared/constants/constants';
+import { NavBarName } from 'shared/enum/enumUA/navigation-bar';
+import { ModalConfirmationType } from 'shared/enum/modal-confirmation';
+import { Direction } from 'shared/models/category.model';
+import { NavigationBarService } from 'shared/services/navigation-bar/navigation-bar.service';
+import { AdminState } from 'shared/store/admin.state';
+import { AddNavPath } from 'shared/store/navigation.actions';
```
doliinyk, line 17: It is OK for replacing such long relative path to shared folder for directly using shared because of paths that specified in tsconfig.json?

doliinyk:
Last commit contains required fix and tested locally, but better need a look for it. Some components have at least two parent routes (places where they can be navigated from) so haven't used route `'../'`.

Also guess it's time to uncomment TODO task in [CreateDirectionComponent at line 57](https://github.com/ita-social-projects/OoS-Frontend/blob/0e3ec6c043ba3697b49c7d427588d4228edd54f9/src/app/shell/admin-tools/data/directions-wrapper/directions/create-direction/create-direction.component.ts#L57) because of addNavPath() method is abstract at the moment in `CreateFormComponent`.

src/app/shell/admin-tools/data/directions-wrapper/directions/create-direction/create-direction.component.ts
humanmantis: It's ok, it just makes importing easier

humanmantis (approved):

PR #2141, opened 2023-07-11
Oliinyk / Tech admin create area admin

doliinyk:
Implemented creating/updating territorial community (area) admin for tech admin, add territorial community admin service and actions; it is able to list, update and delete territorial community admin well.
Also add `parentId` option for Codeficator filtering, refactored Admin store state and actions.


doliinyk:
#2143

Also going to migrate autocomplete form fields of region and territorial community to dropdown selects.

src/app/shared/services/territorial-community-admin/territorial-community-admin.service.ts
```diff
@@ -0,0 +1,71 @@
...
+import { SearchResponse } from '../../models/search.model';
+import { TerritorialCommunityAdmin, TerritorialCommunityAdminParameters } from '../../models/territorialCommunityAdmin.model';
+import { BaseAdminService } from '../base-admin/base-admin';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class TerritorialCommunityAdminService extends BaseAdminService {
+  constructor(protected http: HttpClient) {
+    super(http, 'areaAdmin');
```
humanmantis: Can we use enum `AdminRoles`? territorialCommunityAdmin value is not used in the application, so we can freely change it

humanmantis (approved):

PR #2153, opened 2023-07-17
Bytsa/500 filters for institutions depending on rights

Ultrafiolet4:
Made three dropdowns that have insitutions, regions, areas. Depending on admin role, admin can apply filters which he is allowed. Made reset filter button, refactored code(separated to methods), fixed bug with hovering of translate dropdown options.


src/app/shell/admin-tools/data/provider-list/provider-list.component.html
```diff
@@ -1,17 +1,72 @@
...
+            disableOptionCentering
+            formControlName="region"
+            panelClass="dropdown-panel"
+            placeholder="{{ 'FORMS.PLACEHOLDERS.SELECT_REGION' | translate }}"
+            class="select">
+            <mat-option
+              class="dropdown-option"
+              *ngFor="let region of regions$ | async"
+              [value]="region"
+            >
```
humanmantis: Keep bracket on the same line

src/app/shell/admin-tools/data/provider-list/provider-list.component.scss
```diff
@@ -58,3 +60,91 @@
...
+.filters-area, .region-filters-area {
+  column-gap:5px;
+  margin-right: 12px;
+}
+.reset {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  gap: 0.5rem;
+
```
humanmantis: No need for empty spaces between properties

src/app/shell/admin-tools/data/provider-list/provider-list.component.ts
```diff
@@ -67,19 +107,50 @@ export class ProviderListComponent implements OnInit, OnDestroy {
...
   };
+  public regions$: Observable<Codeficator[]>;
 
-  constructor(private liveAnnouncer: LiveAnnouncer, private store: Store, private matDialog: MatDialog) {}
+  constructor(private liveAnnouncer: LiveAnnouncer,
+              protected route: ActivatedRoute,
+              private store: Store,
+              private matDialog: MatDialog,
+              private formBuilder: FormBuilder) {
+  }
```
humanmantis: It is better to format like this

```
constructor(
  ...,
  ...
) {}

```

src/app/shell/admin-tools/data/provider-list/provider-list.component.ts
```diff
@@ -94,7 +165,49 @@ export class ProviderListComponent implements OnInit, OnDestroy {
...
+      this.selectedAdmin$.subscribe((admin: RegionAdmin) => {
+        this.store
+          .dispatch(new GetCodeficatorById(admin.catottgId))
+          .pipe(
+            takeUntil(this.destroy$),
+            switchMap((state) =>
+              this.store.dispatch(new GetCodeficatorSearch(state.metaDataState.codeficator.region, [CodeficatorCategories.Level1]))
+            )
+          )
+          .subscribe((state) => {
```
humanmantis: Avoid using nested subscribtions. Use switchMap instead

src/app/shell/admin-tools/data/provider-list/provider-list.component.ts
```diff
@@ -103,12 +216,102 @@ export class ProviderListComponent implements OnInit, OnDestroy {
...
+    this.institutionFormControl.valueChanges
+      .pipe(
+        distinctUntilChanged(),
+        startWith(''),
+        skip(1),
+        debounceTime(1000),
+        takeUntil(this.destroy$)
+      )
+      .subscribe((value: string) => {
+        if (value !== '') {
```
humanmantis: Can we replace `if (value !== '')` with `filter(Boolean)` operator in pipe?

src/app/shell/admin-tools/data/provider-list/provider-list.component.ts
```diff
@@ -94,7 +165,49 @@ export class ProviderListComponent implements OnInit, OnDestroy {
...
 
-    this.filterFormControl.valueChanges
+    this.filterGroup = this.formBuilder.group({
+      searchBarFilter: new FormControl(''),
+      institution: new FormControl(''),
+      region: new FormControl(''),
+      area: new FormControl('')
+    });
+
+    if (this.isTechAdmin) {
```
humanmantis, line 179: Perhaps this logic should be made into a separate method (up to 204 row)

src/app/shell/admin-tools/data/provider-list/provider-list.component.ts
```diff
@@ -202,25 +408,86 @@ export class ProviderListComponent implements OnInit, OnDestroy {
...
+  }
+
   ngOnDestroy(): void {
     this.destroy$.next(true);
     this.destroy$.unsubscribe();
     this.store.dispatch(new PopNavPath());
   }
 
+  private checkFiltersParametersForReset() {
+    if (this.providerParameters.searchString !== '' || this.providerParameters.size !== PaginationConstants.TABLE_ITEMS_PER_PAGE
```
humanmantis: This condition looks so scary. Is there any way to optimize it?

humanmantis: At least `this.providerParameters.searchString !== ''` can be just `!this.providerParameters.searchString`

And perhaps it would be good to create several variables for such conditions: `(this.isTechAdmin && (this.providerParameters.catottgId !== '' || this.providerParameters.institutionId !== ''))`

Another way is to think in reverse logic, which may require fewer conditions

humanmantis (changes requested):

src/app/shell/admin-tools/data/provider-list/provider-list.component.ts
Ultrafiolet4: We cant make !this.providerParameters.searchString, because its not working with this condition, but at least I can optimize this condition with getters

humanmantis: Creating getters may be a overkill idea if you only use them in one place. It can be just few `const`.

humanmantis: It still looks pretty scary. 
Do we need this condition at all, because by resetting the filter, we will still set the default parameters? Won't it work if we just modify the `onResetFilters` in this way or similar:
```
const hasFilter = Object.values(this.filterGroup.controls).some((control: AbstractControl) => Boolean(control.value));
if (hasFilter) {
  this.filterGroup.reset({
    searchBarFilter: '',
    institution: '',
    region: '',
    area: ''
  }, { emitEvent: false });

  this.areaFormControl.disable();

  this.store.dispatch(new ClearCodeficatorSearch());

  this.setProviderFilterByDefault();
  this.getProviders();
}
```

humanmantis: I have tested this solution with the Tech Admin and the Ministry Admin, and it works great, when the filter is empty, we don't send unnecessary requests.

humanmantis (changes requested):

src/app/shell/admin-tools/data/provider-list/provider-list.component.ts
Ultrafiolet4: Thanks for your responce and help! This is really great version of this functionality, but I changed it a little bit and now it works perfect and looks much better!

humanmantis (approved):

PR #2174, opened 2023-07-26
Set max length on input in chat and made label that showing current number of symbols and max number

Ultrafiolet4:


humanmantis (changes requested):
Use double quotes in templates

humanmantis (approved):

PR #2176, opened 2023-07-26
Set max number of "TO" age input, refactored quotes and imports

Ultrafiolet4:


humanmantis (changes requested):
Use double quotes in templates

humanmantis (approved):

Issue #2197, opened 2023-08-07
[UI]  "Workshop". Page with workshop details 


OlenaYakubova:
**Environment:** Windows 11 64 bit, Google Chrome 114.0.5735.199
**Reproducible:** always.

**Preconditions**
1. Go to OoS website (e.g. https://oos.dmytrominochkin.cloud/#/) as any role of user
2. Geographic location is selected (e.g. Kyiv)
3. There is at least one registered workshop (Човник) in the selected location

**Step to reproduce**
1. Go to the page with workshop details by clicking the 'Детальніше' link of selected workshop card (Човник)

**Actual result**
1. There is too large space between "Міністерство культури та інформаційної політики України" and "Міністерство молоді та спорту України" hyperlinks 
**Actual result**
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/73597578/b9d765a2-fa44-4862-bd0a-28f69f82feb5)
**Expected result**
 ![image](https://github.com/ita-social-projects/OoS-Frontend/assets/73597578/867849f7-7eff-4db0-9660-e7632045cd86)
2. Next to the workshop rating, there should be a note about the number of people who voted:
**Actual result**
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/73597578/a50ca80e-25c4-4ff1-bceb-ba0480018100)
**Expected result**
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/73597578/4542e718-74bb-4937-a094-d3392250fe39)
3. There is no icon opposite the workshop name:
**Actual result**
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/73597578/68a03812-15b4-4459-993c-301a54b9aca8)
**Expected result**
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/73597578/12518603-d269-4f2c-a4b4-37a7053aa4b8)
4. The title of the section should be "Розклад занять"
**Actual result**
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/73597578/ab7dced0-c1a4-4ceb-92ac-36722f4eef61)
**Expected result**
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/73597578/ef62b881-c4f3-4342-b094-015f8f6586f8)

**User story and test case links**
[Test case][#192](https://jira.softserve.academy/browse/OS-192)

**Labels to be added**
"Bug", Priority ("Medium"), Severity ("Low"), Type "UI"

[...]

PR #2199, opened 2023-08-08
Oliinyk / 2197 Page with workshop details

doliinyk:
#2197


doliinyk:
How do look hyperlinks in footer:
image

Rating with number of ratings:
image

Direction image next to workshop name:
image

Translation of schedule:
image

Important notes:
~~1. Number of ratings displayed only in workshop details, not in list (maybe have to and if is - waiting that field in backend response);~~
2. Direction image is not displayed in details if workshop doesn't have direction

humanmantis (approved):

Issue #2207, opened 2023-08-14
[Personal cabinet] The "Зберегти" button works several times, and returns the browser pages back when you quickly click the "Зберегти" button several times.


Shliakhtych:
**Environment:** Windows 10, 64-bit, Google Chrome 115.0.5790.171,
Windows 10, 64-bit, Microsoft Edge 114.0.1823.67.
**Reproducible:** always.

**Description**

When you quickly press the "Save" button, it works several times and returns the browser pages several pages back.

**Preconditions**

1. Go to (https://oos.dmytrominochkin.cloud/)
2. The user is authorized as a parent.

**Steps to reproduce**

1. Go to the "Особиста інформація" tab.
2. Click on "Редагувати" button.
3. Quickly press the "Зберегти" button several times.

**Actual result**
The information is saved, the user returns several pages back.

**VIDIO - https://app.screencast.com/nBH0eSbUgOkb7**

**Expected result**
The information is saved, the user returns to the "Особиста інформація" page.

**Labels to be added**
"Bug", Priority ("pri: High"), Severity ("severity: High").

PR #2211, opened 2023-08-15
2207 personal cabinet the save button works several times and returns the browser pages back when you quickly click the save button several times

Ultrafiolet4:
#2207


src/app/shell/admin-tools/platform/platform-info/info-edit/info-edit.component.ts
```diff
@@ -149,16 +170,8 @@ export class InfoEditComponent extends CreateFormComponent implements OnInit, On
...
-        platformInfoItemArray.push(new CompanyInformationSectionItem(form.value))
-      );
-
-      const platformInfo = this.editMode
-        ? new CompanyInformation(this.titleFormControl.value, platformInfoItemArray, this.platformInfo.id)
-        : new CompanyInformation(this.titleFormControl.value, platformInfoItemArray);
-
-      this.store.dispatch(new UpdatePlatformInfo(platformInfo, this.platformInfoType));
+      this.isDispatching = true;
+      this.dispatchSubject.next();
```
humanmantis, line 177: I don't think we need a Subject here

We can modify the method like this and it should be the same.
The same in the other case
```
onSubmit(): void {
    if (this.PlatformInfoItemArray.valid && this.titleFormControl.valid) {
      this.isDispatching = true;
      const platformInfoItemArray: CompanyInformationSectionItem[] = [];
      this.PlatformInfoItemArray.controls.forEach((form: FormGroup) =>
        platformInfoItemArray.push(new CompanyInformationSectionItem(form.value))
      );

      const platformInfo = this.editMode
        ? new CompanyInformation(this.titleFormControl.value, platformInfoItemArray, this.platformInfo.id)
        : new CompanyInformation(this.titleFormControl.value, platformInfoItemArray);

      this.store.dispatch(new UpdatePlatformInfo(platformInfo, this.platformInfoType)).subscribe(() => {
        this.isDispatching = false;
      });;
    }
  }
```

humanmantis (changes requested):

src/app/shell/admin-tools/platform/platform-info/info-edit/info-edit.component.ts
Ultrafiolet4: In this case we need Subject here, because of one little bug - when you clicking fast on a start of request, it can make one or 2-3 requests max. (only,  but its enough). It's not good for performance and, for example, user get on main page after updating own info. We need to be sure, so I think, throttleTime will be barrier of making unnecessary requests and isDispatching flag will be main functionality.

humanmantis: I tried it locally, and in this case it works fine. But maybe I missed something. 
We can use the current solution, it's great anyway. 
Just a few notes: 
- create a variable (for example, `defaultThrottleTime`) instead of a magic number
- move the logic from `switchMap` to a separate method so that we can minimize the code in `ngOnInit`

src/app/shell/admin-tools/platform/platform-info/info-edit/info-edit.component.ts
```diff
@@ -54,6 +58,15 @@ export class InfoEditComponent extends CreateFormComponent implements OnInit, On
 
   ngOnInit(): void {
     this.route.params.pipe(takeUntil(this.destroy$)).subscribe((params: Params) => this.setInitialData(params));
+
+    this.dispatchSubject
```
humanmantis, line 62: Can be minimized to
```
 this.dispatchSubject
      .pipe(
        throttleTime(this.defaultThrottleTime),
        switchMap(() => this.updatePlatformInfoInStore())
       )
      .subscribe(() => this.isDispatching = false);
```

humanmantis (changes requested):

humanmantis (approved):

PR #2222, opened 2023-08-29
Oliinyk / Admin block provider number requiring

doliinyk:
Now blocking provider modal requires admin phone number for feedback.
It is how it looks:
image


doliinyk:
Also add displaying admin phone number for blocked provider:
image

doliinyk:
Making it ready-to-review PR because backend improvements are not hard to do

humanmantis:
I don't remember how it was before, but there should probably be some kind of label for textarea before the asterisk

src/app/shared/components/confirmation-modal-window/reason-modal-window/reason-modal-window.component.html
```diff
@@ -3,31 +3,60 @@ <h2 mat-dialog-title>{{ modalTitle | translate }}</h2>
...
-        ng-blur="updateValue()"
-        class="step-textarea"
-        placeholder="{{ 'FORMS.PLACEHOLDERS.MAXIMUM' | translate }} {{ validationConstants.MAX_DESCRIPTION_LENGTH_500 }} {{
-          'FORMS.PLACEHOLDERS.SYMBOLS' | translate
-        }}"
-        [formControl]="ReasonFormControl"
-        appTrimValue
-      ></textarea>
+      <textarea matInput
+                ng-focus="updateValue()"
```
humanmantis: `ng-focus` is used in AngularJS, use `(focus)` instead

humanmantis (changes requested):

doliinyk:
>

doliinyk:
> I don't remember how it was before, but there should probably be some kind of label for textarea before the asterisk

In deploy it is looks the same

humanmantis:
> > I don't remember how it was before, but there should probably be some kind of label for textarea before the asterisk
> 
> In deploy it is looks the same

Understood, but this is an area that could be improved in the future, because the empty line with an asterisk may not be clear to all users. It might be worth adding some kind of label, such as "Reason for blocking" or something similar.

humanmantis (approved):

PR #2248, opened 2023-09-13
Oliinyk / Area admin list region display

doliinyk:
Fixed displaying redundant column in ministry admins list, add region displaying for area admins & refactored


src/app/shared/components/users-list/users-list.component.ts
humanmantis, line 1: Add access modifiers for properties and methods

humanmantis (changes requested):

humanmantis (approved):

Issue #2255, opened 2023-09-22
 [Сhild Information] The "Зберегти" button works several times, and returns the browser pages back when you quickly click the "Зберегти" button several times.


Shliakhtych:
**Environment:** Windows 10, 64-bit, Google Chrome 117.0.5938.88,
Windows 10, 64-bit, Microsoft Edge 114.0.1823.67.
**Reproducible:** always.

**Description**

When you quickly press the "Save" button, it works several times and returns the browser pages several pages back.

**Preconditions**

1. Go to (https://oos.dmytrominochkin.cloud/)
2. The user is authorized as a parent.
3. At least 1 child must be created.

**Steps to reproduce**

1. Go to the "Інформація про дітей" tab.
2. Select a child from the list and Click on the "Редагувати" button.
3. Quickly press the "Зберегти" button several times.

**Actual result**
The information is saved, the user returns several pages back.

VIDEO - https://app.screencast.com/eHZPgprOZmTn6

**Expected result**
The information is saved, the user returns to the "Особиста інформація" page.

**Labels to be added**
"Bug", Priority ("pri: High"), Severity ("severity: High").

Shliakhtych:
The error is still there. The defect continues to reproduce.

Ultrafiolet4:
Bug fixed in #2257 pull request

PR #2257, opened 2023-09-23
Made confirmation modal for child edit save button

Ultrafiolet4:
#2255


src/app/shell/personal-cabinet/parent/create-child/create-child.component.ts
```diff
@@ -198,28 +158,73 @@ export class CreateChildComponent extends CreateFormComponent implements OnInit,
   /**
    * This method create or edit Child and dispatch CreateChild action
    */
-  onSubmit(): void {
-    if (this.ChildrenFormArray.invalid) {
+  public onSubmit(): void {
+    if (this.ChildrenFormArray.invalid && this.isDispatching) {
```
humanmantis: We do not need to do any work during dispatching

humanmantis (changes requested):

src/app/shell/personal-cabinet/parent/create-child/create-child.component.ts
Ultrafiolet4: We discussed a similar problem in another place and decided that this condition should be checked

src/app/shell/personal-cabinet/parent/create-child/create-child.component.ts
```diff
@@ -198,28 +158,75 @@ export class CreateChildComponent extends CreateFormComponent implements OnInit,
...
-  onSubmit(): void {
-    if (this.ChildrenFormArray.invalid) {
-      this.checkValidationChild();
-    } else {
-      const parent = this.store.selectSnapshot<Parent>(RegistrationState.parent);
-      if (this.editMode) {
-        const child: Child = new Child(this.ChildrenFormArray.controls[0].value, parent.id, this.child.id);
-        this.store.dispatch(new UpdateChild(child));
+  public onSubmit(): void {
+    if (this.isDispatching) {
```
humanmantis: `!this.isDispatching` is probably correct version.

humanmantis (changes requested):

humanmantis (approved):

PR #2291, opened 2023-10-19
fixed page loads endlessly

SergeyGSA:


src/app/shared/services/applications/application.service.ts
```diff
@@ -1,23 +1,19 @@
-import { catchError, Observable, tap, throwError } from 'rxjs';
+import { Observable } from 'rxjs';
```
humanmantis: Angular imports should come first

src/app/shared/services/applications/application.service.ts
```diff
@@ -39,12 +35,14 @@ export class ApplicationService {
         params = params.set('SearchString', parameters.searchString);
       }
 
+      if ('size' in parameters && 'from' in parameters) {
```
humanmantis, line 37: Can we check this same as above? (`parameters.size && parameters.from`)

humanmantis (changes requested):

src/app/shared/services/applications/application.service.ts
SergeyGSA: I did it at first but faced with an issue. A parameter From can be equal "0", and this condition returns false but has to return true because this pagination's parameter exists. So, I decided to use this check.





If you tell me better solution, I'll be appreciate.

humanmantis: In this case, we can use the initial solution

humanmantis (approved):

PR #2322, opened 2023-12-20
Oliinyk / Phone number input lib

doliinyk:
Add 3-rd party lib that using libphonenumber by Google to formatting/input international phone numbers.
Removed hardcoded phone country code in template


src/app/shared/components/phone-form-control/phone-form-control.component.ts
humanmantis, line 1: Directly overriding methods is a bad practice. Is it possible to inherit class from the library and override the methods if necessary?
And it is better to implement this component as custom form field (which is probably done in the class from the library, just check if it works fine)
Ref: https://material.angular.io/guide/creating-a-custom-form-field-control

src/app/shared/components/phone-form-control/phone-form-control.component.scss
humanmantis, line 1: Use more specific classes for `::ng-deep` as it will override styles for all elements that have `.mat-menu` or `.country-selector` class

src/app/shared/pipes/phone-transform.pipe.ts
```diff
@@ -1,19 +1,13 @@
 import { Pipe, PipeTransform } from '@angular/core';
-import { Constants } from '../constants/constants';
+import { isValidPhoneNumber, parsePhoneNumber } from 'libphonenumber-js';
 
 @Pipe({
   name: 'phoneTransform'
```
humanmantis: Just name it `phone`

humanmantis (changes requested):

src/app/shared/components/phone-form-control/phone-form-control.component.ts
doliinyk: > Directly overriding methods is a bad practice. Is it possible to inherit class from the library and override the methods if necessary? And it is better to implement this component as custom form field (which is probably done in the class from the library, just check if it works fine) Ref: https://material.angular.io/guide/creating-a-custom-form-field-control

It isn't possible because of library requirements of providing the `Countries` array from the `/lib` path in `/node_modules`, but it doesn't export from the package.
image
Component extending also requires writing `@Component` decorator (because metadata doesn't inherit).
Here is what it looks like:
```ts
@Component({
  selector: 'app-phone-custom',
  template: '',
  providers: [{ provide: CountryCode, useValue: CountryCode }]
})
export class CustomInputComponent extends NgxMatIntlTelInputComponent
```
image

humanmantis: It's sad news, so for now we can leave it at that, as there is no better workaround.

humanmantis (approved):

src/app/shared/components/phone-form-control/phone-form-control.component.ts
```diff
@@ -46,4 +47,12 @@ export class PhoneFormControlComponent implements AfterViewInit, OnDestroy {
...
     this.destroy$.unsubscribe();
   }
+
+  @HostListener('document:keypress')
+  public handleKeyboardEvent() {
+    const countrySearchElement = document.querySelector('.ngx-mat-tel-input-mat-menu-panel .country-search') as HTMLElement;
+    if (countrySearchElement) {
+      setTimeout(() => countrySearchElement.focus(), 200);
+    }
+  }
```
doliinyk: These lines fix the issue of country search becoming unfocusable on slow typing.
Like there:

The approach described [here](https://github.com/tanansatpal/ngx-mat-intl-tel-input/issues/157) is used to handle every `document:keypress` event (it is used whole document event because it doesn't emit in country search but somewhere in menu container). However, you can see a little bit blink of focus events.
There how it looks now:

humanmantis (approved):

PR #2325, opened 2024-01-04
Task/implement history log of parent blocking

SergeyGSA:
- added history log of parents blocking
- reworked filters of history log


src/app/shell/admin-tools/data/history-log/history-log-filters/history-log-filters.component.ts
```diff
@@ -1,41 +1,93 @@
...
-      dateFrom: new FormControl(monthAgoDate),
-      dateTo: new FormControl(currentDate),
-      options: new FormControl('')
-    });
-
     this.filterData.emit(this.filtersForm.value);
   }
+  
+  public ngOnChanges(changes: SimpleChanges): void {
+    this.setBaseFiltersForm();
```
humanmantis, line 29: Do I need to do this every time a change occurs, or only when switching tabs?

src/app/shell/admin-tools/data/history-log/history-log.component.ts
```diff
@@ -163,4 +174,26 @@ export class HistoryLogComponent implements OnInit, OnDestroy {
     Util.setFromPaginationParam(this.filters, this.currentPage, this.totalAmount);
     this.dispatchProperValue(this.tabIndex, this.filters, searchString);
   }
+
+  private initSubscribeOnEachHistory(): void {
```
humanmantis, line 178: It looks like we have multiple subscriptions to perform the same action.
Can we combine them using a merge operator or some other option?

humanmantis (changes requested):

src/app/shell/admin-tools/data/history-log/history-log-filters/history-log-filters.component.ts
SergeyGSA: We need this only when switching tabs, but I can't find better solution. Could you help me with this?

SergeyGSA: We need this only when tabs switching, so I can move this code to the component on higher level. There are mat-tabs in this component and I can observe tabs changing and execute this methods. What do you think?

humanmantis: We can use a setter for input property (tabName in my understanding) to track that the tab has been changed and reset the form only in this case.
E.g: 
```

  @Input public set tabName(newValue) {
    this._tabName = newValue;
    // reset form
    // ...
  }

  public get tabName() {
    return this._tabName;
  }

```

humanmantis (changes requested):

humanmantis (approved):

Issue #2045, opened 2023-03-30
[Technical administrator] Filtering for the selected period is not worked on the "Історія змін" tab.


ViktoriiaHrytchyna:
**Environment:** Windows 10, 64-bit, Google Chrome 107.0.5304.107.
**Reproducible:** always.

**Preconditions**
1.Go to [https://oos.dmytrominochkin.cloud/#](https://oos.dmytrominochkin.cloud/)
2.Log in as SP and change "Підпорядкування" and "Назва закладу"("Email" = '[lidoyo6572@cmeinbox.com](mailto:lidoyo6572@cmeinbox.com)', "password" = 'Qazwsx789!')
3.Log in as Technical administrator ("Email" = '[admin@gmail.com](mailto:admin@gmail.com)', "password" = 'A12345a@')

**Steps to reproduce**
1.Go to "Адміністрування" on the top right corner drop down list.
2.Click on "Історія змін" tab.
3.Pay attention to the "Заклади" page (A list of institutions that have undergone changes is displayed).
4.Select the period from 03.10.2023 to 03.30.2023 in the "Період" field.
5.Click on the "Показати" button.
6.Pay attention to the empty change history on the "Заклади" page of the "Історія змін" tab.
7.Click on the "Скинути фільтри"and after that click on the "Показати" button.
8.Select "Підпорядкування" from the filter drop-down menu and after that click on the "Показати" button.
9.Pay attention to the empty change history on the "Заклади" page of the "Історія змін" tab.
10.Click on the "Скинути фільтри"and after that click on the "Показати" button.
11.Enter alphanumeric and symbolic values ​​in the "Період" field.
12.Click on the "Показати" button.
13.Pay attention that there is no invalid data error.

**Actual result**
Filtering for the selected period is not worked on the "Історія змін" tab. You can enter alphanumeric and symbolic values ​​in the period field.
![Снимок экрана 2023-03-30 135727](https://user-images.githubusercontent.com/115226563/228823756-547107ac-9181-46d5-9fdd-94384d17de97.jpg)
![Screenshot_40](https://user-images.githubusercontent.com/115226563/228823864-1415e5b1-e3ac-4b70-ad7a-565edec06b5c.jpg)
![Screenshot_1](https://user-images.githubusercontent.com/115226563/228824151-76bf7d58-4611-4317-942a-6d01506c9a77.jpg)

**Expected result**
[...]

PR #2369, opened 2024-02-05
Morozov restrict enter alpha letters 2045

vladmorozov2:
Issue: (#2045 )

Added new directive which checks if input is date symbols and applied to datapicker


src/app/shared/directives/date-symbols-checker.directive.ts
humanmantis, line 1: Please reuse the KeyFilter directive, just pass it the appropriate RegExp from constants.

humanmantis (changes requested):

vladmorozov2:
Made git pull and just used keyFilter with appropriate regexp.
Please review again

humanmantis (approved):

Issue #2012, opened 2023-02-06
[City-filter] - no-city placeholder should not be cut off


litvinets:
Not city placeholder should be either full-width, to part of the text should be on the next line
![image](https://user-images.githubusercontent.com/27493959/216913252-8b1fa405-fcd9-4e9a-a7b8-a0689e3f4f59.png)

PR #2370, opened 2024-02-05
made placeholder bigger #2012

vladmorozov2:
Issue: (#2012)

made search autocomplete bigger
image


src/app/shared/components/filters-list/city-filter/city-filter.component.scss
```diff
@@ -10,14 +10,17 @@
...
 }
+
 .input-filter {
   text-transform: capitalize;
   font-family: 'Innerspace';
   color: #000000;
   font-size: 1.125rem;
   line-height: 1.25rem;
   font-weight: 500;
+  min-width: 290px;
```
humanmantis: Please check if it works with different screen resolutions

vladmorozov2: It now looks good on any devices.
This selector wont affect any potential new mat-autocomplete because it have id selector which unique and due to scope of css component files

humanmantis: cdk-overlay-id can be different, so don't use it.
You can provide class as input for autocomplete or use panelWidth option.
Ref: https://v14.material.angular.io/components/autocomplete/api

humanmantis (changes requested):
Do not use cdk-overlay-id class.
Please check docs for the autocomplete component and specify class/width for panel

src/app/shared/components/filters-list/city-filter/city-filter.component.scss
vladmorozov2: Is it okay [panelWidth]="300"?

humanmantis: Yes, its fine

humanmantis (approved):

PR #2385, opened 2024-02-19
Horeslavskyi/fixing data filter in history of changes

SergeyGSA:
Setup filters in history of changes


src/app/shared/enum/history.log.ts
```diff
@@ -6,6 +6,8 @@ export enum HistoryLogTypes {
 }
 
 export enum FilterOptions {
+  dateFrom = 'dateFrom',
```
humanmantis: It is better to keep the same PaskalCase for all options.
value can be in CamelCase

src/app/shell/admin-tools/data/history-log/history-log-filters/history-log-filters.component.ts
```diff
@@ -1,100 +1,150 @@
...
+  }
+
+  private setTimePeriodEqualToWholeDay(dateFrom: Date, dateTo: Date): FilterData {
+    this.setCustomTimeInDate(dateFrom, 0, 0, 0);
+    this.setCustomTimeInDate(dateTo, 23, 59, 59);
+
+    return this.setTimeDependsOnTimezone(dateFrom, dateTo);
+  }
+
+  private addFormControlForFiltersForm(formControlName: string, additionalFormControlName?: string): void {
```
humanmantis: You can pass an array of control names, so we don't need to add an additional parameter if we need more controls in the future

humanmantis (changes requested):

src/app/shell/admin-tools/data/history-log/history-log-filters/history-log-filters.component.ts
SergeyGSA: I don't understand exactly. I have to assign a new value to the variables on 143 and 146 lines from this array, because they are used in the layout. How can I do it without editing this method in the future? I don't get it.

This is the case  how I see it, but I'm not satisfied of this

<pre>
  private addFormControlForFiltersForm(formControlsNames: string[]): void {
    formControlsNames.forEach((formControlName: string, index: number) => {
      this.filtersForm.addControl(formControlName, new FormControl(''));

      if (index === 0) {
        this.formControlName = formControlName;
      } else if (index === 1) {
        this.additionalFormControlName = formControlName;
      }
    });
  }
   </pre>
   
    What do you think?

humanmantis: In my understanding, it should create a kind of config that can be used in a template.

As a result, we should get something like this 
```
{ 
  formControlName,
  options,
  order/id/etc.
} 
```

As a result, we don't need many variables to display drop-down lists in the template, just one ngFor loop will be enough.
You may need to create a different config that matches form control name and drop-down options.

As a result, we get something like this

```
 private addFormControlForFiltersForm(formControlNames: string[]): void {
    formControlNames.forEach((controlName: string) => {
      this.filtersForm.addControl(controlName, new FormControl(''));
      this.config.push({ controlName, options: this.dropdownOptionsConfig[controlName] });
    });
  }
  ```

humanmantis (approved):

PR #2406, opened 2024-03-06
fixed moving text in application-card

MarianS23:
Added the empty block to save the position of the text when the user is blocked and the button panel has been disappeared


src/app/shell/personal-cabinet/shared-cabinet/applications/application-card/application-card.component.html
```diff
@@ -68,32 +68,40 @@ <h5 *ngIf="userRole === role.provider">
...
     <ng-container *ngIf="isMobileView; then actionsBlock"></ng-container>
   </mat-card-content>
-  <mat-card-content class="card-block card-block__buttons" *ngIf="userRole === role.provider && !application.isBlockedByProvider">
-    <ng-container *ngIf="application.workshop.competitiveSelection; else approveButtonBlock">
-      <ng-container *ngIf="application.status === applicationStatuses.Pending">
-        <button class="btn" (click)="acceptForSelection.emit(application)" [ngClass]="isApproveBtnHidden && 'hidden'">
-          <p>{{ 'BUTTONS.ACCEPT_TO_COMPETITIVE_SELECTION' | translate }}</p>
-        </button>
+
+  <ng-container *ngIf="userRole === role.provider && !application.isBlockedByProvider; else elseTemplate">
```
humanmantis: It is better to call the template by a more specific name, `elseTemplate` is too general.

src/app/shell/personal-cabinet/shared-cabinet/applications/application-card/application-card.component.html
```diff
@@ -68,32 +68,40 @@ <h5 *ngIf="userRole === role.provider">
...
+    <mat-card-content class="card-block card-block__buttons">
+      <ng-container *ngIf="application.workshop.competitiveSelection; else approveButtonBlock">
+        <ng-container *ngIf="application.status === applicationStatuses.Pending">
+          <button class="btn" (click)="acceptForSelection.emit(application)" [ngClass]="isApproveBtnHidden && 'hidden'">
```
humanmantis: In the case when we set only one class, it is better to do it like this:
`[class.hidden]="condition"`

humanmantis: Change it in other places as well

humanmantis (changes requested):

Issue #2345, opened 2024-01-18
[Error text] Change the standard wording of the error to a more customized one when all places in the workshop are occupied on the "Applications" page.


Shliakhtych:
**Environment:** Windows 10, 64-bit, Google Chrome 107.0.5304.107.
**Reproducible:** always.

**Preconditions**

1. Go to [https://oos.dmytrominochkin.cloud/#](https://oos.dmytrominochkin.cloud/)
2. Log in as a Service Provider.
3. The service provider must have a workshop where all seats are occupied.
4. The service provider must have at least 1 unaccepted application in a workshop in which all places are occupied.

**Steps to reproduce**

1. Go to the "Applications" page.
2. Find an unaccepted application in a workshop where there is no more space.
3. Click on the "Зарахувати" button.
4. Pay attention to the message that appears.

**Actual result**
You need to change the wording of the error message from "На жаль, виникла помилка" to "Вибачте, місць у гуртку більше немає."
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/134736243/07c555b2-a571-490e-bab8-f9c89a74a4a5)

**Expected result**
The error wording is displayed correctly.

**Labels to be added**
"Bug", Priority ("priority: Low "), Severity ("severity: Trivial"), Type ("UI").

PR #2407, opened 2024-03-07
Oliinyk / 2345 Workshop seats occupied api error

doliinyk:
#2345

Refactored message-bar & some application components & add TODO workshop lack seats api error type


src/app/shell/personal-cabinet/shared-cabinet/applications/application-card/application-card.component.html
```diff
@@ -1,15 +1,13 @@
-<mat-card [ngClass]="application.status === applicationStatuses.Pending ? 'card card-new' : 'card'">
-  <mat-card-content [ngClass]="userRole === role.parent ? 'card-block card-block__parent' : 'card-block'">
+<mat-card [ngClass]="application.status === ApplicationStatuses.Pending ? 'card card-new' : 'card'">
```
humanmantis, line 1: class `card` is same for both conditions.
We can move it to class property, so condition will looks better.
In that case we can do `[class.card-new]="condition"` because there will be only one additional class.

humanmantis: I see that you have only changed the formatting in this file, so this comment may be optional. 
But I prefer to make these changes across the file

humanmantis (approved):

src/app/shell/personal-cabinet/shared-cabinet/applications/application-card/application-card.component.html
doliinyk: Unfortunately, didn't noticed the comment and merged already, but it'll come with next PR of map fixes

Issue #2388, opened 2024-02-23
[Service provider cabinet] The 'i' button is not displayed near the 'Форма власності' field when the SP edits the profile information


OMGaranina:
**Environment:** OS Windows 11 Pro, Version 22H2; Browser: Chrome, Version 121.0.6167.18
**Reproducible:** always
**Build found:** commit 9f52723

**Preconditions**
1. The site https://oos.dmytrominochkin.cloud/#/ is opened
2. User is logged in as a Service Provider

**Steps to reproduce**
1. Click on the user`s name in the header
2. Choose the 'Інформація про заклад' from the drop down list
3. Click on the 'Редагувати' button
4. Observe the 'Форма власності' section

**Actual result**
The 'i' button with additional information is not displayed near the 'Форма влісності' section
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/95030828/84fed394-c2fd-41eb-b0e5-c106e60efd21)

**Expected result**
The field “Форма власності" is read only. The 'i' button is near the field. The additional information pop up appears when clicking on the 'i' button.

**User story and test case links**
[User story #992](https://github.com/ita-social-projects/OoS-Backend/issues/992)

PR #2436, opened 2024-03-22
Issue#2388 service provider cabinet the i button is not displayed near the ownership field when the sp edits the profile information

SergeyGSA:
https://github.com/ita-social-projects/OoS-Frontend/issues/2388

- Replaced all mat-menus with "i" button with one common component
- Added "i" button near the Ownership field when SP edits the profile information

Medium screens:
normal

Small screens:
small


src/app/shared/components/info-menu/info-menu.component.ts
```diff
@@ -0,0 +1,28 @@
...
+import { InfoMenuType } from 'shared/enum/info-menu-type';
+
+@Component({
+  selector: 'app-info-menu',
+  templateUrl: './info-menu.component.html',
+  styleUrls: ['./info-menu.component.scss']
+})
+export class InfoMenuComponent {
+  @Input() public isBigIcon: boolean;
+  @Input() public typeOfContent: InfoMenuType;
```
humanmantis: Can you set a default value for the type?
We can also rename it to just `type`.

src/app/shared/enum/info-menu-type.ts
```diff
@@ -0,0 +1,4 @@
+export enum InfoMenuType {
+  ComponentEntity,
```
humanmantis: Why not just `Component`?

humanmantis: It is better to set a value for enum

humanmantis (changes requested):

humanmantis (approved):

PR #2381, opened 2024-02-13
made transition to /personal-cabinet/provider/info after registation

vladmorozov2:
made transition to /personal-cabinet/provider/info after registation

PR #2443, opened 2024-03-27
Oliinyk / Unauthorized user error

doliinyk:
https://github.com/ita-social-projects/OoS-Frontend/pull/2381

Fixed error coming when checking auth if user unauthorized.
image

The problem was lying behind the requirement of returning a stream from the action.

### DIFF
--- a/src/app/shared/store/registration.state.ts
+++ b/src/app/shared/store/registration.state.ts
@@ -131,7 +131,7 @@ export class RegistrationState {
   }
 
   @Action(Login)
-  login(_: StateContext<RegistrationStateModel>, { payload }: Login): void {
+  login(_ctx: StateContext<RegistrationStateModel>, { payload }: Login): void {
     const configIdOrNull = null;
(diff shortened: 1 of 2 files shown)


src/app/shared/store/registration.state.ts
```diff
@@ -131,7 +131,7 @@ export class RegistrationState {
   }
 
   @Action(Login)
-  login({}: StateContext<RegistrationStateModel>, { payload }: Login): void {
+  login(_: StateContext<RegistrationStateModel>, { payload }: Login): void {
```
doliinyk: Also need to confirm the approach. The underscore is used to avoid SonarLint unexpected object pattern code smell and marking the argument as ignored; note that injecting StateContext here is necessary for method to perform.

humanmantis: The underscore at the beginning of the unsed parameter is okay, but it's better to add more sense, maybe do something like `_context` or `_ctx`.

humanmantis (approved):

PR #2466, opened 2024-04-10
Romanchak / Creation and update Admin

Kazumen:
The AdminRole enum was changed to lowercase, and the bugs that were associated with this action were fixed


src/app/shell/admin-tools/data/admins/create-admin/create-admin.guard.ts
humanmantis, line 1: I was thinking about possible modifications to the guard. Maybe we can make the solution easier. `сreateRole` is not a `Role`, it's `AdminRoles`. Maybe we can just replace the check in switch...case from `Role.ministryAdmin` to `AdminRoles.ministryAdmin` and so on

humanmantis (changes requested):

src/app/shell/admin-tools/data/admins/create-admin/create-admin.guard.ts
Kazumen: Yes, it looks much simpler, but to me it looks the same as toLowerCase()

humanmantis: It seems like the same thing, but we actually use the correct types instead of workarounds

humanmantis (approved):

Issue #2497, opened 2024-04-18
[UI; SP cabinet] No clarification of the "Blocked" status in info tooltip


OlenaYakubova:
**Environment:** Windows 11, 64 bit
Google Chrome 124.0.6367.61
Reproducible: always.

**Preconditions**
1. Go to https://oos.dmytrominochkin.cloud/
2. Log in as Service provider

**Steps to reproduce**
1. Click on the arrow in right top corner of page
2. Select "Заявки" line from drop down list
3. Click on "i" icon (info-menu) befor "всі гуртки" filter

**Actual result**
There is no clarification of the "Blocked" status in the Application Statuses.
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/73597578/b809c4e1-fc19-47dc-931c-1fa798ec4b2a)

**Expected result**
Clarification of the "Blocked" status is in the Application Statuses in Service Provider cabinet 
Should be "Заблоковано" - "Користувача заблоковано" like:
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/73597578/a00a3b5a-1a09-4908-822a-6f33784cf3e9)

**User story and test case links**
(Figma)[https://www.figma.com/file/YRYP4zAAchOyYMgdoeP782/MoE-%7C-Dev-Team?type=design&node-id=0-1&mode=design&t=Bfz9yxJ5gU6N1TlB-0]

**Labels to be added**
"Bug", Priority ("Low"), Severity ("Low"), Type ("UI").

AkunaPatlata:
@OlenaYakubova I requested access to figma. Now waiting for you to open it to me:)

PR #2511, opened 2024-04-29
Dvorak/2497 clarification of the blocked status in info tooltip added

AkunaPatlata:
#2497 
Fixed blocked status in info tooltip showing up for every user. Now it is appearing only for providers.


src/app/shared/enum/statuses.ts
```diff
@@ -33,6 +33,17 @@ export enum ApplicationStatuses {
   Banned = 'Banned'
 }
 
+// Statuses for Applications
+export enum ApplicationStatuses {
```
humanmantis: Instead of copy-pasting all values, we can use Exclude utility
https://www.typescriptlang.org/docs/handbook/utility-types.html#excludeuniontype-excludedmembers

src/app/shared/components/status-info-card/status-info-card.component.html
humanmantis, line 1: We have almost the same template for both options. It is better to unify it.

src/app/shared/components/status-info-card/status-info-card.component.ts
```diff
@@ -1,19 +1,24 @@
...
 import { Application } from '../../models/application.model';
+import { Role } from '../../enum/role';
 
 @Component({
   selector: 'app-status-info-card',
   templateUrl: './status-info-card.component.html',
   styleUrls: ['./status-info-card.component.scss']
 })
 export class StatusInfoCardComponent {
+  @Input() public role: Role;
```
humanmantis: Make this input a setter so that we can keep one property for the status ( `public status: ApplicationStatuses | ApplicationProviderStatuses`) and set it based on the role and the template will remain as simple as before

humanmantis (changes requested):

src/app/shared/components/status-info-card/status-info-card.component.html
```diff
@@ -1,13 +1,15 @@
...
-      /></span>
-    </div>
-    <div class="application-info-description" fxFlex="70" fxLayoutAlign="space-between start">
-      <span class="{{ status }}">{{ applicationStatusDescription[status.key] | translate }}<br /></span>
+  <div>
+    <div *ngFor="let status of statuses | keyvalue" class="status-group" fxLayout="row" fxLayoutAlign="start stretch">
+      <div class="application-info-title" fxFlex="30" fxLayoutAlign="space-between start">
+        <span class="{{ status.key }}"
+          ><i class="{{ applicationIcons[status.key] }}"></i>{{ statusTitles[status.key] | translate }}<br
+        /></span>
```
humanmantis: formatting looks broken

src/app/shared/components/status-info-card/status-info-card.component.ts
```diff
@@ -3,17 +3,24 @@ import { ApplicationIcons } from '../../enum/applications';
...
 import { Application } from '../../models/application.model';
+import { Role } from '../../enum/role';
 
 @Component({
   selector: 'app-status-info-card',
   templateUrl: './status-info-card.component.html',
   styleUrls: ['./status-info-card.component.scss']
 })
 export class StatusInfoCardComponent {
+  @Input() public application: Application = null;
```
humanmantis: I don't see it used anywhere. You can remove it

src/app/shared/components/status-info-card/status-info-card.component.ts
```diff
@@ -3,17 +3,24 @@ import { ApplicationIcons } from '../../enum/applications';
...
+  @Input() public role: Role;
+
   public readonly statusTitles = ApplicationTitles;
   public readonly applicationStatusDescription = ApplicationStatusDescription;
   public readonly applicationIcons = ApplicationIcons;
-  public readonly statuses = ApplicationStatuses;
-
-  @Input() application: Application = null;
+  public readonly Role = Role;
+  public get statuses(): any {
```
humanmantis: This is not exactly what I thinking.
In the example below, we will update the statuses only when the role is set.
```
private _role: Role;
  public get role(): Role {
    return this._role;
  }
  @Input()
  public set role(value: Role) {
    this._role = value;
    this.setStatuses(); // can be same code, as in lines 22-24
  }
```

src/app/shared/components/status-info-card/status-info-card.component.ts
```diff
@@ -3,17 +3,24 @@ import { ApplicationIcons } from '../../enum/applications';
...
+    return this.role === Role.provider
+      ? ApplicationStatuses
```
humanmantis: Can we use array of enum values instead? 
`Object.values(ApplicationStatuses)`

and when role is not provider
`Object.values(ApplicationStatuses).filter((status: ApplicationStatuses) => status !== ApplicationStatuses.Banned)`

humanmantis (changes requested):

src/app/shared/enum/statuses.ts
```diff
@@ -21,7 +21,7 @@ export enum UserStatuses {
   Blocked = 'Blocked'
 }
 
-// Statuses for Applications
+// Statuses for Provider Applications
```
humanmantis, line 24: revert change

humanmantis (approved):

Issue #2494, opened 2024-04-17
[Create/Edit Achievement] Achievement title text is not transferred correctly when expanding the full achievement title


Shliakhtych:
**Environment:** Windows 10, 64-bit, Google Chrome 123.0.6312.58.
**Reproducible:** always.

**Preconditions**

1. Go to https://oos.dmytrominochkin.cloud/;
2. Authorize as a service provider;
3. The service provider must have at least 1 workshop.

**Steps to reproduce**

1. Go to my workshops page;
2. Select a workshop and click the More Details button;
3. Click on the Add Achievements button;
4. Click on the Add achievement or Edit achievement button;
5. Open a drop-down list with the names of achievements;
6. Place the cursor on the achievement;
7. Pay attention to the displayed text.

**Actual result**
The text of the full title of the workshop achievements is not transferred correctly when the cursor is hovered over.
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/134736243/edda3696-9404-49de-9e40-01006c4c07a9)
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/134736243/e00c5e94-abaa-479f-a8f3-494b1b5efb3d)
![image](https://github.com/ita-social-projects/OoS-Frontend/assets/134736243/b81efabb-e51b-4a51-9a22-b666fd319687)

**Expected result**
The full name of the workshop achievement is displayed correctly.

**Labels to be added**
"Bug", Priority ("Priority: Medium"), Severity ("Severity: Major"), Type ("UI").

PR #2519, opened 2024-05-03
Oliinyk / 2494 Achievement tooltip word break

doliinyk:
#2494

Changed style for whole tooltips word break property as overflow-wrap.
Used `overflow-wrap: break-word` property since `word-break: break-word` is deprecated one.
Style is global and applied for every tooltip, for example city-filter.
image


doliinyk:
Please, do merge after approval.

humanmantis (approved):

Issue #2404, opened 2024-03-05
[District selection] "After scrolling down with selected desctription whole pages roll down"


WVNTICXRV:
**Environment:** Windows 10, 64-bit, Google Chrome 121.0.6167.86.
**Reproducible:** always

**Preconditions**

The site 'Позашкілля' https://oos.dmytrominochkin.cloud/ is opened
The user is logged in as a service provider or parent

**Steps to reproduce**

Click on 'Обрати місто';
Scroll down the district list;
Select the district description;
Scroll down while 
the cursor is on the description text;

https://github.com/ita-social-projects/OoS-Frontend/assets/110986476/f6e967e6-c14e-4821-a6ff-8336f4be05e5

**Actual result**
The main page scrolled down. The district selection window doesn't move

**Expected result**
After scrolling down district window rolls down and other districts are seen

**Labels to be added**
"Bug", Priority ("pri: Low "), Severity ("severity: Low "), Type ("UI, "Functional").

PR #2520, opened 2024-05-09
#2404 scrolling whole page when scroll in district select list

MarianS23:
added custom mat tooltip default options ,tooltip disappear after cursor leave trigger zone.


src/app/app.module.ts
```diff
@@ -95,7 +97,8 @@ registerLocaleData(localeUk);
       provide: HTTP_INTERCEPTORS,
       useClass: ErrorHandleInterceptor,
       multi: true
-    }
+    },
+    { provide: MAT_TOOLTIP_DEFAULT_OPTIONS, useValue: CustomMatTooltipDefaultOptions }
```
humanmantis: Do not provide this options globally

humanmantis (changes requested):

src/app/app.module.ts
MarianS23: done

src/app/shared/constants/constants.ts
```diff
@@ -148,3 +149,12 @@ export class ModeConstants {
 }
 
 export const EMPTY_RESULT = { totalAmount: 0, entities: [] };
+
+export const CustomMatTooltipDefaultOptions: MatTooltipDefaultOptions = {
```
humanmantis: Give it a more specific name. "Custom" is a very general word. Maybe something like "NoInteractTooltipOptions".

humanmantis (changes requested):

humanmantis (approved):

Issue #2536, opened 2024-05-19
[User personal cabinet] Dropdown does not fit the size of the text and information icon is displayed far from child's name


SyrvatkaTetiana:
**Environment:** Mobile version IOS 17.4.1, browser: Google Chrome 
**Reproducible:** always
**Build found:**  "commit [99a4163](https://github.com/ita-social-projects/OoS-Frontend/commit/99a4163b918b677cb5dd046ae3237ebf0bb1c675)"

**Preconditions**
1. The site https://oos.dmytrominochkin.cloud/#/ is opened
2. User is registered as Parent

**Steps to reproduce**
1. Go to 'User personal cabinet'
2. Click on 'Заявки'

**Actual result**
1. 'Всі діти' dropdown does not fit the size of the text
2. Information icon is displayed in the middle of the page, far from child's name
3. 'Прийнято до конкурсного відбору' status is shifted one character to the right under workshop

**Expected result**
1. 'Всі діти' dropdown fits the size of the text in it
2. Information icon should be displayed near child's name
3. 'Прийнято до конкурсного відбору' status is aligned under workshop

<img width="279" alt="image" src="https://github.com/ita-social-projects/OoS-Frontend/assets/151456891/ad1d5248-a3f0-4e82-9238-1fb939b9c8f3">

PR #2539, opened 2024-05-20
Dvorak / 2536 Fixed some styles in user personal cabinet for mobile screens.

AkunaPatlata:
#2536 

- Fixed dropdown. Now text fits in perfectly. 

- Also information icon is now positioned propperly. 

- Also fixed some statuses that were shifted one character to the right under workshop

### DIFF
--- a/src/app/shell/personal-cabinet/shared-cabinet/applications/applications.component.scss
+++ b/src/app/shell/personal-cabinet/shared-cabinet/applications/applications.component.scss
@@ -15,6 +15,10 @@
   margin-bottom: 1rem;
 }
 
+::ng-deep .mat-form-field-infix {
+  width: 110px !important;
+}
+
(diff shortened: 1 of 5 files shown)


src/app/shell/personal-cabinet/shared-cabinet/applications/applications.component.scss
```diff
@@ -15,6 +15,10 @@
   margin-bottom: 1rem;
 }
 
+::ng-deep .mat-form-field-infix {
```
humanmantis: This will change the styles for all elements with the `.mat-form-field-infix` class in the entire application with ng-deep.
Add an additional class that will only match the linked form fields.

humanmantis (changes requested):

humanmantis (approved):

PR #2540, opened 2024-05-21
Smyk / Lowered filter fields on Administator's "History of change" tab 

AndriiSmyk:
Changed field stylesheets so that they are on the same level.
Changed datepicker styles so the content is in the middle of the container.
image

### DIFF
--- a/src/app/shell/admin-tools/data/history-log/history-log-filters/history-log-filters.component.scss
+++ b/src/app/shell/admin-tools/data/history-log/history-log-filters/history-log-filters.component.scss
@@ -93,8 +93,10 @@
   padding: 0;
 }
 
-:host ::ng-deep .mat-form-field-infix {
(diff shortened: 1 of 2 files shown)


src/app/shell/admin-tools/data/history-log/history-log-filters/history-log-filters.component.scss
```diff
@@ -92,6 +93,10 @@
   padding: 0;
 }
 
+:host ::ng-deep .mat-form-field-infix {
```
humanmantis: I understand that this should basically apply to all form fields in the component, but just in case, add an additional class, something like `no-padding-form-field` or similar

humanmantis (changes requested):

humanmantis (approved):

Issue #2583, opened 2024-06-19
[New workshop application form] The button '+Додати ще дитину' should be disabled if the parent already has 20 children


OMGaranina:
**Environment:** OS Windows 11Pro Version 23H2, Browser Chrome Version 125.0.6422.114
**Reproducible:** always
**Build found:** commit [5e3f71a](https://github.com/ita-social-projects/OoS-Frontend/commit/5e3f71a75710f389bfb9d0a6a512ad8afa40152b)

**Preconditions**
1. The site 'Позашкілля' https://oos.dmytrominochkin.cloud  is opened
2. The user is logged in as a Parent
3. The parent has 20 registered children

**Steps to reproduce**
1. Click on the 'Детальніше' button on the workshop card
2. Click on the 'Подати заявку' button
3. Hover over the '+Додати ще одну дитину' button
4. Click on the '+Додати ще одну дитину' button

**Actual result**
The '+Додати ще одну дитину' button is active. The arrow cursor changed to the pointer. After clicking on the button the 'Додати дані про дитину' form is opened.

https://github.com/ita-social-projects/OoS-Frontend/assets/95030828/eff40f7d-bc7e-4f32-a26b-d5ad473b0f1e

**Expected result**
The '+Додати ще одну дитину' button is disabled. The arrow cursor is not changing to the pointer. After clicking on the button '+Додати ще одну дитину' any actions are not initiated. Near the '+Додати ще одну дитину' button add an information button 'i'. After clicking on the 'i' button the pop-up message appeared 'Можливо додати не більш, ніж 20 дітей'
<img width="950" alt="більше20" src="https://github.com/ita-social-projects/OoS-Frontend/assets/95030828/b0d7f2ee-19a4-49c3-88ef-002170cea19e">

**User story and test case links**
[User story#1012](https://github.com/ita-social-projects/OoS-Backend/issues/1012)

Shliakhtych:
The bug has been fixed.
![image](https://github.com/user-attachments/assets/7592b388-4840-43a2-b4b5-d5ccfaa3b309)

PR #2585, opened 2024-06-22
Smyk / 2583 The button '+Додати ще дитину' should be disabled if the parent already has 20 children

AndriiSmyk:
#2583 
Disabled 'Add another child' button when the parent already has 20 children.
Added information button with pop-up message when the parent already has 20 children.

### DIFF
--- a/src/app/shell/personal-cabinet/parent/create-application/create-application.component.html
+++ b/src/app/shell/personal-cabinet/parent/create-application/create-application.component.html
@@ -35,12 +35,12 @@
 
             <p class="background-text" fxLayout="row" fxLayoutAlign="space-around center">
               {{ 'NO_CHILD_FOUND' | translate }}
-              <a [routerLink]="!has20Children ? ['/create-child', ModeConstants.NEW] : []" [queryParams]="!has20Children ? { workshopId: workshop?.id } : {}" [class.disabled-anchor]="has20Children">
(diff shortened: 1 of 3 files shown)


src/app/shell/personal-cabinet/parent/create-application/create-application.component.html
humanmantis, line 1: I think we can make a property that will show that the child limit is reached and reuse it inside the template instead of checking the number of children 5 times

humanmantis (changes requested):

src/app/shell/personal-cabinet/parent/create-application/create-application.component.ts
```diff
@@ -99,6 +100,7 @@ export class CreateApplicationComponent implements OnInit, OnDestroy {
     this.children$.pipe(filter(Boolean), takeUntil(this.destroy$)).subscribe((children: SearchResponse<Child[]>) => {
       this.parentCard = children.entities.find((child: Child) => child.isParent);
       this.children = children.entities.filter((child: Child) => !child.isParent);
+      this.has20Children = this.children.length >= 20;
```
humanmantis: A magic number in the property name is not the best option.
It is better to name it something like `isChildLimitReached` and create a constant MAX_CHILD or similar

humanmantis (changes requested):

humanmantis (approved):

Issue #2573, opened 2024-06-12
[Search page] The ‘List’ page doesn’t scroll up after clicking on the ‘Переглянути всі’ button on the Main page.


oksanakushnir:
**Environment:** Windows 11 Version 22H2, Chrome Version 125.0.6422.142
**Reproducible:** always
**Build found:**  "commit fdcaee0"

**Preconditions**
1.	The site https://oos.dmytrominochkin.cloud/#/ is opened

**Steps to reproduce**
1.	Scroll to the ‘Найпопулярніші гуртки’ section
2.	Click on the ‘Переглянути всі’ button

**Actual result**
The page doesn’t scroll up to the beginning of the section.

https://github.com/ita-social-projects/OoS-Frontend/assets/130609090/5b33ae3b-ee83-4a7b-8d12-fdc2de491016

**Expected result**
The page should be pre-scrolled to the begin.

**User story and test case links**
User story #166

**Labels to be added**
"Bug", Priority ("priority: Low "), Severity ("severity: Minor"), Type ("UI").

Shliakhtych:
The bug has been fixed.

PR #2588, opened 2024-06-25
Shutov / #2573 [ Search page] Added setting to scroll up after switching between routes

MarianS23:
set scrollPositionRestoration app-routing module


MarianS23:
@humanmantis i don't know ,do we need such behavior in entire app?

humanmantis (approved):

humanmantis:
> @humanmantis i don't know ,do we need such behavior in entire app?

For me, this looks like the expected behavior on all pages

Issue #2505, opened 2024-04-21
Replacing legacy Angular Material components with the new MDC-based components


SergeyGSA:
### Replacing legacy Angular Material components with the new MDC-based components
The migration of Angular Material version from 14 to 15 was done in pull request https://github.com/ita-social-projects/OoS-Frontend/pull/2457.

The official documentation of migration is provided in the link below
https://v16.material.angular.io/guide/mdc-migration

### The list of the legacy components that should be migrated:  
- [x] legacy-core (MatLegacyOption as well)

- [x] legacy-radio (https://v16.material.angular.io/guide/mdc-migration#radio)
- [x] legacy-progress-bar (https://v16.material.angular.io/guide/mdc-migration#progress-bar)
- [x] legacy-menu (https://v16.material.angular.io/guide/mdc-migration#menu)
- [x] legacy-chips (https://v16.material.angular.io/guide/mdc-migration#chips)
- [x] legacy-card (https://v16.material.angular.io/guide/mdc-migration#card)
- [x] legacy-button (https://v16.material.angular.io/guide/mdc-migration#button)
- [x] legacy-tooltip (https://v16.material.angular.io/guide/mdc-migration#tooltip)
- [x] legacy-table (https://v16.material.angular.io/guide/mdc-migration#table)
- [x] legacy-tabs (https://v16.material.angular.io/guide/mdc-migration#tabs)
- [x] legacy-snack-bar (https://v16.material.angular.io/guide/mdc-migration#snack-bar)
- [x] legacy-autocomplete (https://v16.material.angular.io/guide/mdc-migration#autocomplete)
- [x] legacy-checkbox (https://v16.material.angular.io/guide/mdc-migration#checkbox)
- [x] legacy-input (https://v16.material.angular.io/guide/mdc-migration#input)
- [x] legacy-form-field (https://v16.material.angular.io/guide/mdc-migration#form-field)
- [x] legacy-dialog (https://v16.material.angular.io/guide/mdc-migration#dialog)
- [x] legacy-select (https://v16.material.angular.io/guide/mdc-migration#select)

### To avoid confusion in the future, I strongly suggest specifying the component from the above list where changes have been made and your pull request. For example: 

![comment_example](https://github.com/ita-social-projects/OoS-Frontend/assets/67392473/80966c17-f2b7-47a1-9ea5-d57bb2e032f7)

doliinyk:
[...]

PR #2592, opened 2024-06-30
Oliinyk / 2505 MDC core, progress-bar, option migration

doliinyk:
#2505

Mat-Legacy-Option wasn't migrated indeed since it's provided by Mat-Legacy-Select now, so it's just test/minor code changes without styles.


doliinyk:
It is not possible to migrate mat-option since it depends in mat-select.

Issue #2581, opened 2024-06-18
[Main page] The invalid characters can be entered in the Search field.


oksanakushnir:
**Environment:** Windows 11 Version 22H2, Chrome Version 125.0.6422.142
**Reproducible:** always
**Build found:** "commit 5e3f71a"

**Preconditions**
1.	The site https://oos.dmytrominochkin.cloud/#/ is opened.

**Steps to reproduce**
1.	Type invalid characters in the Search field (e.g.  和平).
2.	Click on the Search icon.

**Actual result**
The invalid characters can be entered in the field.

**Expected result**
The invalid characters should not be entered in the field. Error message should be appeared.

**User story and test case links**
User story https://github.com/ita-social-projects/OoS-Backend/issues/572

**Labels to be added**
"Bug", Priority ("Medium"), Severity ("Medium"), Type ("Functional")

PR #2593, opened 2024-07-02
Smyk / 2581 The invalid characters can be entered in the Search field

AndriiSmyk:
#2581 

Restricted input of invalid characters in search field.
Created error message when user inputs invalid character.


humanmantis:
Vissually, it looks fine, but why are you choose to use custom directive intead of validator for the field?
I don't think we will use it anywhere else

AndriiSmyk:
> Vissually, it looks fine, but why are you choose to use custom directive intead of validator for the field? I don't think we will use it anywhere else

In my oppinion, it can be used in create workshop form for restricting user to input invalid symbols. Even if it isn't, this directive may come in handy somewhere else.

src/app/shared/directives/search-field-validation.directive.ts
```diff
@@ -0,0 +1,23 @@
+import { Directive, ElementRef, EventEmitter, HostListener, Output } from '@angular/core';
+
+@Directive({
+  selector: '[appSearchFieldValidation]'
+})
+export class SearchFieldValidationDirective {
```
humanmantis: _Left a comment here to keep everything in one thread_

My main point why this is not the best solution for us is that it is better to use common practices for reactive forms. In our application, we usually use validators to check if the value is correct, which is a straightforward, so I don't think we should change it to any other approach.

src/app/shared/components/filters-list/searchbar/searchbar.component.ts
```diff
@@ -78,6 +84,16 @@ export class SearchbarComponent implements OnInit, OnDestroy {
     this.performSearch();
   }
 
+  public handleInvalidCharacter(value: string): void {
+    const validValue = value?.replace(/[^A-Za-zА-Яа-я0-9`.,№"']/g, '');
+    if (validValue !== value) {
+      this.searchValueFormControl.setValue(validValue);
+      this.invalidCharacterDetected.emit();
```
humanmantis, line 91: Can we use app-validation-hint instead of triggering an error to the parent component?

AndriiSmyk: I don't think so, because when we use app-validation-hint in this component it appears directly under search input and causes everything inside this component to shift.

humanmantis: I don't agree with doing different types of validation for different form fields. We should use a consistent approach throughout the application to ensure the same experience and interface. You can try adjust search bar size, leaving a space for the error message there

humanmantis (approved):

PR #2596, opened 2024-07-08
Oliinyk / Lint fixes

doliinyk:


doliinyk:
Please, ignore SonarCloud fail since it's only whitespace/access modifier refactoring in purpose to get rid of lint warnings.

.lintstagedrc
```diff
@@ -0,0 +1,3 @@
+{
+  "*.{ts,scss,html}": ["yarn lint"]
```
humanmantis: `yarn lint:fix` may be the best option for us, because it applies all the fixes on the fly. But there may still be problems with template formatting.

doliinyk: Just tried hook with suggested script - it modifies around 700 files without any changes, looks like line ending fixes. Also trying to fix some mock error - for example, adding semicolon in the end of line. But it still can't fix even .ts prettier formatting issues.

doliinyk: Ok, it looks fine, started it one more time again and got better results. Guess we can deal with it

humanmantis (approved):

Issue #2615, opened 2024-07-29
[Workshop/Create] Required field missed error message during workshop creation.


oksanakushnir:
**Environment:** Windows 11 Version 22H2, Chrome Version 125.0.6422.142
**Reproducible:** Always
**Build found:** "commit eddcd03"

**Preconditions**
1.	The site https://oos.dmytrominochkin.cloud/#/ is opened
2.	Provider is registered.

**Steps to reproduce**
1.	Open the provider information.
2.	Click on the “Мої гуртки” tab.
3.	Click on the “Додати гурток” button.
4.	Scroll to the “Розклад занять” section.
5.	Select few days in toggle.
6.	Enter start time, leave end time unfilled.

**Actual result**
The required end time field does not display an error message.
![2024-07-29_14-10-22](https://github.com/user-attachments/assets/18f95cf5-529d-4277-8872-70abd913b153)

**Expected result**
The required end time field should display an error message.

**Labels to be added**
"Bug", Priority ("Medium"), Severity ("Medium"), Type ("Functional")

Shliakhtych:
The bug has been fixed.
![image](https://github.com/user-attachments/assets/9d8de229-2449-43bf-8e97-d49477dc91fe)

PR #2618, opened 2024-08-01
Smyk / 2615 Required field missed error message during workshop creation

AndriiSmyk:
#2615 

Refactored validation hint component to accept as input either FormControl or FormGroup

### DIFF
--- a/src/app/shared/components/validation-hint/validation-hint.component.ts
+++ b/src/app/shared/components/validation-hint/validation-hint.component.ts
@@ -1,5 +1,6 @@
 import { ChangeDetectorRef, Component, EventEmitter, Input, OnChanges, OnDestroy, OnInit, SimpleChanges } from '@angular/core';
-import { FormControl, ValidationErrors } from '@angular/forms';
+import { FormControl, FormGroup, ValidationErrors } from '@angular/forms';
+import { control } from 'leaflet';
 import { Subject } from 'rxjs';
 import { debounceTime, takeUntil } from 'rxjs/operators';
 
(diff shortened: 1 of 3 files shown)


src/app/shared/components/validation-hint/validation-hint.component.ts
```diff
@@ -51,30 +51,45 @@ export class ValidationHintComponent implements OnInit, OnDestroy, OnChanges {
   constructor(private cdr: ChangeDetectorRef) {}
 
   public ngOnInit(): void {
-    this.validationFormControl.statusChanges.pipe(debounceTime(200), takeUntil(this.destroy$)).subscribe(() => {
-      const errors = this.validationFormControl.errors;
+    if (this.validationFormControl instanceof FormControl) {
```
humanmantis: Can we rewrite this part in the way shown below?

```
this.validationFormControl.statusChanges.pipe(debounceTime(200), takeUntil(this.destroy$)).subscribe(() => {
  if (this.validationFormControl?.controls.length) {
    Object.keys(this.validationFormControl.controls).forEach(() => { ... })
  } else {
    this.updateValidationState(this.validationFormControl as FormControl);
  }
});
```

humanmantis (changes requested):

src/app/shared/components/validation-hint/validation-hint.component.ts
AndriiSmyk: We can't use this condition:
`if (this.validationFormControl?.controls.length)`
because FormGroup.controls doesn't have the `length` property.
However, putting the condtitional statement inside of `subscribe()` sounds great.

humanmantis (approved):

PR #2626, opened 2024-08-24
Oliinyk / lint-staged fix

doliinyk:
- expanded .gitignore files related to extensions & package manager
- changed configuration for lint-staged:
1. Removed scss for lint - since we're not applying lint rules for it
2. Separated lint and prettier use for file extensions
3. Used eslint instead of ng lint (that uses angular eslint builder) in lint-staged - because it wasn't correctly received staged files
4. Used prettier manually in lint-staged because it tried to fix all files, not only staged

That reduces time for pre-commit hook to be executed and helps us to avoid unnecessary template changes by hook.

cc @humanmantis


.lintstagedrc
```diff
@@ -1,3 +1,8 @@
 {
-  "*.{ts,scss,html}": ["yarn lint:fix"]
+  "*.{ts,html}": [
+    "eslint --fix"
```
humanmantis: It may be better to create commands for it in the package json

humanmantis (approved):

PR #2649, opened 2024-11-07
Oliinyk / Token PersonalInfo read replacement & Removing Subrole concept & Employee role

doliinyk:
This PR aims on removing Subrole concept, which was invented in purpose to differ provider users, for example:
- none (provider)
- provider deputy
- provider admin

It's decided to remove this concept and migrate them to Role (only provider was known as Role).
Also, role, that was read from auth token, here's got from PersonalInfo endpoint because token payload will be encrypted due to its sensitive content.

Per 9/12 there's also fixes for new Employee role.

@humanmantis let's ignore current coverage issues since they're mostly detected as renamed files.


src/app/shared/enum/role.ts
```diff
@@ -2,24 +2,21 @@ export enum Role {
...
-  ProviderDeputy = 'ProviderDeputy',
-  ProviderAdmin = 'ProviderAdmin'
-}
-
 export enum EntityType {
   provider = 'provider',
   workshop = 'workshop',
-  ProviderAdmin = 'provideradmin',
-  ProviderDeputy = 'provider',
+  providerDeputy = 'provider',
```
humanmantis: I would like to avoid using the same values for enums.
This was before your changes, but could you please check if it's not difficult to remove this

src/app/shared/enum/role.ts
```diff
@@ -28,3 +25,5 @@ export enum UserTabParams {
   parent,
   child
 }
+
+export const ProviderRoles = [Role.provider, Role.providerDeputy, Role.employee];
```
humanmantis: We only use it in util, so we can probably remove it from here

humanmantis (changes requested):

src/app/shared/enum/role.ts
doliinyk: Possibly, in the past this enum was dedicated to be used in filtering ratings by type. However, in our project it's always `workshop` type is used and for provider `rating` is always returned within the provider info.
Backend also uses only `workshop` for rating https://github.com/ita-social-projects/OoS-Backend/blob/87d225785d61b617b627b958179417313ab63ef6/OutOfSchool/OutOfSchool.WebApi/Controllers/V1/RatingController.cs#L108
Though, I will remove this option

humanmantis (approved):

PR #2657, opened 2024-12-01
Litviniuk/replace flex layout with tailwind

darryyna:
all flexLayout usages were replaced with Tailwind CSS


package.json
```diff
@@ -71,6 +71,7 @@
     "@types/node": "^18.11.9",
     "@typescript-eslint/eslint-plugin": "7.5.0",
     "@typescript-eslint/parser": "7.5.0",
+    "autoprefixer": "^10.4.20",
```
humanmantis: Are autoprefix and postcss required? I don't see that you include them in the plugins, maybe we can remove them?

humanmantis: Remove @angular/flex-layout when done

src/app/shared/components/filters-list/working-hours/working-hours.component.scss
```diff
@@ -12,6 +12,11 @@
 }
 .wrapper {
   margin-top: 16px;
+  display: flex;
```
humanmantis: Why didn't you use tailwind classes in this case?

src/app/shared/components/status-info-card/status-info-card.component.scss
```diff
@@ -26,16 +26,23 @@
 
 .application-info-description,
 .application-info-title {
-  flex: 1;
   max-width: max-content;
   line-height: 17px;
   color: black;
+  place-content: flex-start space-between;
```
humanmantis: Why didn't you use tailwind classes in this case?

src/app/shell/result/result.component.scss
```diff
@@ -29,7 +29,24 @@
     z-index: 1001;
   }
 }
-
+@media (max-width: 1280px) {
+  .view-control-panel {
+    display: flex;
```
humanmantis: Why didn't you use tailwind classes in this case?

humanmantis (changes requested):

src/app/shared/components/status-info-card/status-info-card.component.scss
darryyna: because there are the same properties for both elements and there is no reason to dublicate it

package.json
humanmantis: I don't see the removed package in yarn.lock file. Did you use `yarn remove <package>`?

src/app/shared/components/status-info-card/status-info-card.component.scss
humanmantis: Yes, but you can just use the ready-made tailwind classes, so it shouldn't be a problem.

darryyna: i made that change in previous commit as you said

humanmantis (approved):

DmyMi:
comment for future duplicate - we had to revert this pr :)

PR #2667, opened 2024-12-12
Liuboff/History of search queries and ability to delete previous search result

Liuboff:
_**It was implemented the Business Rules**_
1. The system allows the user to enter a maximum of 64 characters. If the allowed characters in the search bar are exceeded, the system stops entering characters further.
2. The history of old queries should be stored in the list of old queries under the search bar from newest to oldest. The maximum number of queries in the list is 10 queries. The user should be able to delete a query/s by clicking the cross on the right side of the old query.

_**For now in plans to investigate how to perform the Bonus AC**_
1. Query autocomplete: when entering a query, the system offers
autocomplete options that help the user find
the correct terms or the most popular search queries.
2. Search without spelling errors: the system should offer correct
results even if there are spelling errors in the query
(for example, "drwin" should display results for "drawing").


humanmantis:
Please sync branch with develop, this PR shows many unrelated changes
You can also mark a PR as Draft if it is not yet ready for review

PR #2670, opened 2024-12-16
Liuboff/History of search queries and ability to delete previous search result

Liuboff:
It was implemented the Business Rules:
1. The system allows the user to enter a maximum of 64 characters. If the allowed characters in the search bar are exceeded, the system stops entering characters further.
2. The history of old queries should be stored in the list of old queries under the search bar from newest to oldest. The maximum number of queries in the list is 10 queries. The user should be able to delete a query/s by clicking the cross on the right side of the old query.

The next part of task is to investigate and perform the Bonus AC:
- Search without spelling errors: the system should offer correct results even if there are spelling errors in the query (for example, "drwin" should display results for "drawing").
- Query autocomplete: when entering a query, the system offers autocomplete options that help the user find the correct terms or the most popular search queries.


Liuboff:
Coverage - 100%
Reliability - Isn't possible to properly use and implement a 'onKeyPress|onKeyDown|onKeyUp' attribute to <mat-icon>.
(I've tried to reproduce this in application without success. 
When I press Tab, the focus moves from the previous search result to the next element after search. 
Accordingly, the close button cannot be handled by pressing the keyboard keys. 
So I decided not to implement it.)

image

src/app/shared/components/filters-list/searchbar/searchbar.component.html
```diff
@@ -12,8 +12,13 @@
       [matAutocomplete]="autoSearch"
       (keyup.enter)="onValueEnter()" />
     <mat-autocomplete #autoSearch="matAutocomplete" class="search-input" panelClass="dropdown-panel" (optionSelected)="onValueSelect()">
-      <mat-option class="dropdown-option option" *ngFor="let result of filteredResults" [value]="result">
-        {{ result }}
+      <mat-option class="dropdown-option option search-dropdown-option" *ngFor="let result of filteredResults" [value]="result">
+        <span class="option-value">{{ result }}</span>
+        <mat-icon
```
humanmantis, line 17: One possible way to solve the reliability problem is to use a button instead of an icon, try using mat-icon-button insted.
If this doesn't help, I'm fine with leaving it as it is.

src/app/shared/components/filters-list/searchbar/searchbar.component.ts
```diff
@@ -96,6 +96,13 @@ export class SearchbarComponent implements OnInit, OnDestroy {
     }
   }
 
+  public onDeletePreviousSearchValue(value: string, event: Event): void {
+    event.stopPropagation();
+    this.previousResults = this.previousResults.filter((result: string) => result !== value);
+    localStorage.setItem('previousResults', JSON.stringify(this.previousResults));
```
humanmantis: [optional].
I like to have everything under control in one place, so it's better to move it to the NGXS store (we already store part of the store in the browser's SessionStorage). 
We can leave it for now, and in the future we can finalize

src/app/shared/components/filters-list/searchbar/searchbar.component.ts
```diff
@@ -106,14 +113,17 @@ export class SearchbarComponent implements OnInit, OnDestroy {
...
     this.previousResults = this.getPreviousResults();
 
-    if (this.searchedText && !this.previousResults.includes(this.searchedText)) {
-      if (this.previousResults.length > 4) {
-        this.previousResults.shift();
+    const normalizedText = this.searchedText?.trim().toLowerCase();
+    const normalizedResults = this.previousResults.map((result) => result.trim().toLowerCase());
+
+    if (normalizedText && !normalizedResults.includes(normalizedText)) {
+      if (this.previousResults.length >= 10) {
```
humanmantis: Create constant for `10`

humanmantis (changes requested):

src/app/shared/components/filters-list/searchbar/searchbar.component.ts
Liuboff: Done

src/app/shared/components/filters-list/searchbar/searchbar.component.html
Liuboff: Even with mat-icon-button pressing Tab to select the previous search result (and its close button) causes all search results to close.
So we leave it.

src/app/shared/components/filters-list/searchbar/searchbar.component.ts
Liuboff: It is need to check moving functionality to NGXS store.

src/app/shared/components/filters-list/searchbar/searchbar.component.ts
```diff
@@ -41,6 +43,12 @@ export class SearchbarComponent implements OnInit, OnDestroy {
   ) {}
 
   public ngOnInit(): void {
+    this.store.dispatch(new LoadPreviousResults());
+
+    this.previousResults$.pipe(takeUntil(this.destroy$)).subscribe((results: string[]) => {
+      this.previousResults = results;
```
humanmantis: Where do we use this variable now?

src/app/shared/store/filter.state.ts
```diff
@@ -244,6 +253,35 @@ export class FilterState {
...
+
+  @Action(AddPreviousResult)
+  addPreviousResult(ctx: StateContext<FilterStateModel>, { result }: AddPreviousResult): void {
+    const state = ctx.getState();
+    const normalizedResult = result.trim().toLowerCase();
+    const normalizedResults = state.previousResults.map((res) => res.trim().toLowerCase());
+
+    if (normalizedResult && !normalizedResults.includes(normalizedResult)) {
+      const updatedResults = [result, ...state.previousResults.slice(0, Constants.MAX_PREVIOUS_SEARCH_RESULTS - 1)];
+      localStorage.setItem('previousResults', JSON.stringify(updatedResults));
```
humanmantis: Ideally, we should avoid using localStorage directly and store the value in state.
You can have a look at https://www.ngxs.io/v3.8/plugins/storage

We probably won't be able to store part of the state with another storage engine, but maybe we can move it to meta-data.state or something similar, and then we can move other localStorage usages to that state.

humanmantis (changes requested):

src/app/shared/components/filters-list/searchbar/searchbar.component.ts
Liuboff: We don't use. Removed

src/app/shared/store/filter.state.ts
Liuboff: Move localStorage to meta-data.state

humanmantis:
Sync with develop, please 🥺

src/app/shared/store/filter.state.ts
```diff
@@ -244,6 +253,30 @@ export class FilterState {
     patchState({ searchQuery: payload, from: 0 });
   }
 
+  @Action(LoadPreviousResults)
+  loadPreviousResults(ctx: StateContext<FilterStateModel>): void {
```
humanmantis: We can remove this action

src/app/app.module.ts
```diff
@@ -65,8 +67,16 @@ registerLocaleData(localeUk);
     ]),
 
     NgxsStoragePluginModule.forRoot({
-      key: AppState,
-      storage: StorageOption.SessionStorage
+      key: [
+        {
+          key: AppState,
+          engine: SessionStorageEngine
```
humanmantis: Do we need to create custom engines? Is it possible to reuse LOCAL_STORAGE_ENGINE and SESSION_STORAGE_ENGINE from '@ngxs/storage-plugin'?

humanmantis (changes requested):

src/app/shared/store/filter.state.ts
Liuboff: Done

src/app/app.module.ts
Liuboff: Yes, it's possible. Done.

humanmantis (approved):

PR #2672, opened 2024-12-18
Yakymiv/dark theme toggle

YakymivY:


src/app/header/header.component.scss
```diff
@@ -214,6 +218,89 @@
   border-radius: 15px !important;
 }
 
+::ng-deep .light-theme .mat-mdc-slide-toggle,
```
humanmantis: Can we move the theme switcher to a separate component?

src/app/app.component.ts
```diff
@@ -41,14 +47,26 @@ export class AppComponent implements OnInit, OnDestroy {
     this.router.canceledNavigationResolution = 'computed';
     this.store.dispatch([new CheckAuth(), new GetFeaturesList()]);
     this.isWindowMobile(window);
+
+    const savedTheme = localStorage.getItem('preferred-theme');
+    if (savedTheme) {
+      this.isDark = savedTheme === 'dark';
+    } else {
+      this.isDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
```
humanmantis: Use the injection token to manipulate the window. I know there are a few other places where we have the same problem, but we can fix them later

Take a look at: https://www.npmjs.com/package/ngx-window-token

humanmantis (changes requested):
As I understand it, customization of theme colors will be done in separate PRs?
We should probably avoid using custom colors and try to use the most relevant material variables instead

YakymivY:
To enable dark mode correctly we need to rewrite styles in bunch of components, as some of them are hardcoded or need adjustments. So it is another task

src/app/app.component.ts
```diff
@@ -28,9 +30,15 @@ export class AppComponent implements OnInit, OnDestroy {
     private store: Store,
     private translateService: TranslateService,
     private dateAdapter: DateAdapter<Date>,
-    private router: Router
+    private router: Router,
+    @Inject(WINDOW) private _window: Window
```
humanmantis: `_window` => `window`

src/app/app.component.ts
```diff
@@ -41,14 +49,26 @@ export class AppComponent implements OnInit, OnDestroy {
...
+
+    const savedTheme = localStorage.getItem('preferred-theme');
+    if (savedTheme) {
+      this.isDark = savedTheme === 'dark';
+    } else {
+      this.isDark = this._window.matchMedia('(prefers-color-scheme: dark)').matches;
+    }
+  }
+
+  public switchMode(isDarkMode: boolean): void {
```
humanmantis: move all logic inside switcher component

humanmantis (changes requested):

src/app/app.component.ts
```diff
@@ -41,14 +49,25 @@ export class AppComponent implements OnInit, OnDestroy {
     this.router.canceledNavigationResolution = 'computed';
     this.store.dispatch([new CheckAuth(), new GetFeaturesList()]);
     this.isWindowMobile(window);
+
+    // const savedTheme = localStorage.getItem('preferred-theme');
```
humanmantis: Do we need this code?

src/app/app.component.ts
```diff
@@ -29,8 +31,14 @@ export class AppComponent implements OnInit, OnDestroy {
     private translateService: TranslateService,
     private dateAdapter: DateAdapter<Date>,
     private router: Router
+    // @Inject(WINDOW) private window: Window
   ) {}
 
+  @HostBinding('class')
+  public get themeMode(): string {
+    return this.isDark ? 'dark-theme' : 'light-theme';
```
humanmantis: Can we move this part to onToggleChange as well? You can set this class for body element or for the entire document

humanmantis (changes requested):

humanmantis (approved):

PR #2694, opened 2025-01-13
Zhylka/entering workshop hours validation fix

MaxZhylka:
The server does not save time in the format "9:30", so the regular expression was updated to only allow the "09:30" format. The time range validator was previously overriding other errors in the controls, which could result in invalid values without showing any errors.

 ":" autocomplete feature was added. But there's an issue with its implementation: the ":" is currently inserted after the third digit is entered. The user might not realize that the autocomplete feature exists because the third input symbol is typically a ":" entered manually. Changing the condition to insert ":" after the second character would make it impossible for the user to delete it.


src/app/shell/personal-cabinet/provider/create-workshop/create-about-form/working-hours-form-wrapper/working-hours-form/working-hours-form.component.ts
```diff
@@ -52,7 +52,7 @@ export class WorkingHoursFormComponent implements OnInit, OnDestroy {
 
     this.startTimeFormControl.valueChanges.pipe(takeUntil(this.destroy$)).subscribe((value) => {
       this.startTimeFormControl.setValue(this.validateTimeInput(value), { emitEvent: false });
-      if (value) {
+      if (value && !this.startTimeFormControl.errors?.invalidTimeFormat) {
```
humanmantis: Use `this.startTimeFormControl.hasError(...)` instead

humanmantis (approved):

Issue #3035, opened 2025-07-16
[Workshop] [Contacts] The address block disappears after clicking the Continue button if invalid data is entered


Shliakhtych:
**Environment:** Windows 11 Pro, Google Chrome 131.0.6778.86 (64-bit)
**Reproducible:** always.
**Build found:** last commit c94ed28399c7255692e829190caae6d97bd82278

**Steps to reproduce**

1. Go to https://pozashkillia-test.iea.gov.ua/#/
2. Log in as a SP;
3. Go to my workshops page;
4. Сlick the create button;
5. Fill in all required fields and go to the contacts tab;
6. Enter invalid data into any field and collapse the address block;
7. Click the Next button.

**Actual result**
After clicking the Next button, the entire address block disappears from the page, and the provider is unable to proceed with creating the workshop.

https://github.com/user-attachments/assets/db2a8ab0-726a-4a34-a51f-31e2512c6d18

**Expected result**
The service provider can collapse and expand the address block at any time, and it does not disappear from the page.

PR #3082, opened 2025-08-25
Oliinyk/Supporting edge case when stepping next in workshop form while address invalid and collapsed

doliinyk:
https://github.com/ita-social-projects/OoS-Frontend/issues/3035

## Summary by CodeRabbit

* **Bug Fixes**
  * Form validation now scrolls to and focuses only the first actually visible invalid field (ignores hidden or zero-size elements and those inside hidden parents), reducing unexpected jumps and improving accessibility.
  * When a submit handler is provided, the submission path is used instead of advancing the stepper.

* **UI**
  * Removed a stray character in a button to prevent an unintended visible text node.

* **Tests**
  * Added edge-case coverage for element visibility and expanded submit vs. next-step behavior tests.

### DIFF
--- a/src/app/shared/directives/stepper/stepper.directive.ts
+++ b/src/app/shared/directives/stepper/stepper.directive.ts
@@ -59,9 +59,35 @@ export class StepperDirective {
 
         this.store.dispatch(new ShowMessageBar({ message, type: 'error' }));
 
-        invalidFields[0].scrollIntoView({ behavior: 'smooth', block: 'center' });
-        (invalidFields[0] as HTMLElement).focus({ preventScroll: true });
+        const firstVisibleInvalidField = Array.from(invalidFields).find((field) => this.isElementVisible(field as HTMLElement));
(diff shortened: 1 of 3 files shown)


src/app/shared/directives/stepper/stepper.directive.ts
```diff
@@ -59,9 +59,35 @@ export class StepperDirective {
...
+        if (firstVisibleInvalidField) {
+          firstVisibleInvalidField.scrollIntoView({ behavior: 'smooth', block: 'center' });
+          (firstVisibleInvalidField as HTMLElement).focus({ preventScroll: true });
+        }
       }
     });
   }
+
+  private isElementVisible(element: HTMLElement): boolean {
+    const computedStyle = window.getComputedStyle(element);
```
humanmantis: Please use WINDOW injection token instead of working directly with the Window object.

humanmantis (changes requested):

humanmantis (approved):

PR #3123, opened 2025-09-24
Oliinyk / 3117 Create about form age accepts decimal values

doliinyk:
#3117 

## Summary by CodeRabbit

- **Bug Fixes**
  - Age display improved: handles null/undefined/NaN as missing, returns non-negative whole numbers (floored) and truncates overly long values for consistent display.

- **Chores**
  - Public API: age-formatting now accepts and returns null for missing values to ensure consistent handling.

### DIFF
--- a/src/app/shared/utils/utils.ts
+++ b/src/app/shared/utils/utils.ts
@@ -115,7 +115,7 @@ export class Util {
    */
   public static updateStructureForTheTable(users: Child[]): UsersTableData[] {
     const updatedUsers = [];
-    users.forEach((user) => {
+    for (const user of users) {
       updatedUsers.push({
         id: user.id,
         pib: this.getFullName(user),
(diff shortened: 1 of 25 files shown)


src/app/shared/utils/utils.ts
```diff
@@ -115,7 +115,7 @@ export class Util {
    */
   public static updateStructureForTheTable(users: Child[]): UsersTableData[] {
     const updatedUsers = [];
-    users.forEach((user) => {
+    for (const user of users) {
```
humanmantis: Why did you change it to “for of”?

doliinyk: Because of Sonar linting

humanmantis: I don't think it's necessary in this case. Internal methods are more optimized and provide better performance.

humanmantis (approved):

